PR #4, opened 2021-07-27
Make Kotlin compiler produce Java 11 bytecode

kevincianfarini:


mcchang:
Is there more context as to why we need to do this?

mcchang (approved):

kevincianfarini:
@mcchang The Kotlin compiler will, by default, only generate Java 8 bytecode. 

> Kotlin lets you choose the version of JVM for execution. By default, the Kotlin/JVM compiler produces Java 8 compatible bytecode. If you want to make use of optimizations available in newer versions of Java, you can explicitly specify the target Java version from 9 to 16. Note that in this case the resulting bytecode might not run on lower versions.

There have been a lot of performance improvements to JVM bytecode since Java 8. Most notable, `invokedynamic` is more flexible and be used for things like string concatenation.

Issue #5, opened 2021-07-28
Return real LCV scores for representatives


kevincianfarini:
# Querying our database for representative LCV scores 

Objectively speaking, asking LCV to expose an API for scoring representatives is ideal. Since the frequency of LCV score updates is about every 6 months to once yearly, it's pretty low priority to have this information be totally dynamic. Therefore, we'll own these values in our database and update them as needed until we determine it's necessary for LCV to provide us this data dynamically. 

I think the big challenge here will be how we reference representatives. Representatives _do_ have a unique `bioguide`, but I'm failing to find a good source to get that ID for everyone. We're currently using the Google Civic Information API, [which doesn't currently provide a bioguide for representatives](https://groups.google.com/g/google-civicinfo-api/c/TG8b_TnPhII).

We _could_ switch back to Geocodio which provides bioguies in congressional searches. That would make this lookup easy. _However_, Geocodio is not a free API service like Google's civic API. We only get 2500 free requests per day. Ideally, Google's civic API starts providing bioguides. The linked thread above is fairly recent, but for some reason I don't have permissions to comment my interest for that information. 

Unless we have a better option, I propose that we switch back to Geocodio and structure the table as follows. 

| bioguide_id | lcv_score | score_year | 
| --------------- | -------------| ---------------- |
| M001200 | 90 | 2020 |
| M001200 | 88 | -1| 

LCV has previous scores as well as "lifetime" scores, which will be represented as `-1`. We should create a composite, unique key on `(bioguide_id, score_year)`. 

Let me know if you think anything needs to change or if this is good enough to move forward!

kevincianfarini:
Something else to note. Geocodio is substantially slower than Google's civic API. Google requests usually take about 100ms and geocodio takes about 1100ms.

mcchang:
Using bioguide_id seems fine for now.  I've used [Open Civic Identifier](https://opencivicdata.readthedocs.io/en/latest/ocdids.html) person ids ([example](https://github.com/openstates/people/blob/main/data/ak/legislature/Adam-Wool-094c65c5-5c69-4eee-b462-368aadcc6260.yml#L1)) in another project as a unique key for representatives if we want to look for an alternative at some point.  

It sounds like the problem we're trying to solve is how to get from an address to unique identifiers for reps. Is the LCV score dataset keyed by `bioguide_id` currently?

[...]

PR #6, opened 2021-07-29
Migrate to Geocodio

kevincianfarini:
Geocod.io provides bioguide IDs for legislators
which can be used to make cross-service relations
for legislators.

This will be useful when querying LCV scores and
providing image URLs.

Related: #5


backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/service/GeocodioService.kt
```diff
@@ -0,0 +1,14 @@
+package com.climatechangemakers.act.feature.findlegislator.service
+
+import com.climatechangemakers.act.feature.findlegislator.model.GeocodioApiResult
+import retrofit2.http.GET
+import retrofit2.http.Query
+
+interface GeocodioService {
+
+  @GET("geocode?fields=cd")
```
kevincianfarini, line 9: remove this hard coded query param

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/service/GeocodioService.kt
```diff
@@ -0,0 +1,14 @@
...
+  suspend fun getLegislators(
```
mcchang, line 10: feels like this function could be called `geocode`? The only thing that is "legislator" related is the default argument for `fields`

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
```diff
@@ -14,39 +15,43 @@ import kotlinx.coroutines.coroutineScope
...
-      civicResponse.offices.forEach { office ->
-        if (index in office.legislatorIndices) return@Array office.role
-      }
-      error("No legislator role for index $index!")
-    }
+    val geoCodioResponse: GeocodioApiResult = geocodioService.getLegislators(query = request.queryString)
+
+    val geocodioLegislators = geoCodioResponse.results
+      .flatMap { it.fields.congressionalDistricts }
+      .first()
```
mcchang, line 27: should we handle the case that there's no congressional districts for some reason gracefully?

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
```diff
@@ -14,39 +15,43 @@ import kotlinx.coroutines.coroutineScope
...
 
 private val GetLegislatorsRequest.queryString: String get() = "$streetAddress, $city $state $postalCode"
 
-private fun GoogleCivicLegislator.toDomainLegislator(lcvScores: List<LcvScore>, role: LegislatorRole) = Legislator(
-  name = name,
-  role = role,
-  siteUrl = urls.first(),
-  phone = phoneNumbers.first(),
-  imageUrl = photoUrl,
+private val GeocodioLegislator.fullName get() = "${bio.firstName} ${bio.lastName}"
```
mcchang, line 45: would this getter be useful as a public getter defined on `GeocodioLegislator`?

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
```diff
@@ -14,39 +15,43 @@ import kotlinx.coroutines.coroutineScope
...
+  siteUrl = contactInfo.siteUrl,
+  phone = contactInfo.phone,
+  imageUrl = references.getLegislatorImage(),
   lcvScores = lcvScores,
-)
\ No newline at end of file
+)
+
+private fun LegislatorReferences.getLegislatorImage() =
+  "https://bioguide.congress.gov/bioguide/photo/${bioguide[0]}/$bioguide.jpg"
```
mcchang, line 57: this seems useful as a getter on `GeocodioLegislator` for convenience

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
```diff
@@ -14,39 +15,43 @@ import kotlinx.coroutines.coroutineScope
 import javax.inject.Inject
 
 class LegislatorsManager @Inject constructor(
-  private val civicService: GoogleCivicInformationService,
+  private val geocodioService: GeocodioService,
   private val lcvScoreManager: LcvScoreManager,
 ) {
 
   suspend fun getLegislators(request: GetLegislatorsRequest): List<Legislator> = coroutineScope {
```
mcchang, line 22: Would be nice to make this function and its associated request more specific.  We might want to add other ways to look up legislators later.
```suggestion
  suspend fun getLegislatorsByAddress(request: GetLegislatorsByAddressRequest): List<Legislator> = coroutineScope {
```

backend/src/main/kotlin/com/climatechangemakers/act/di/ServiceModule.kt
```diff
@@ -15,18 +15,18 @@ import retrofit2.Retrofit
     val originalRequest = chain.request()
     val originalUrl = originalRequest.url()
     val newUrl = originalUrl.newBuilder()
-      .addQueryParameter("key", requireNotNull(System.getenv("GOOGLE_CIVIC_API_KEY")))
+      .addQueryParameter("api_key", requireNotNull(System.getenv("GEOCODIO_API_KEY")))
```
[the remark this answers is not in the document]
kevincianfarini: I can scope this client so that you need to manually request the geocodio one. Eg. 

```kotlin
fun providesRetrofit(@Geocodio client: OkHttpClient) = ...
```

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/service/GeocodioService.kt
kevincianfarini: agreed.

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
kevincianfarini: I don't think so. This is purely a network model. In our Domain `Legislator` module, we have the full name field. I'm not opposed to storing first and last on the domain and then exposing the fullname as a calculated property, though.

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
kevincianfarini: Do you have an example of what you're talking about? 

Generally, I don't try to over-engineer for a future which might not manifest. We could always provide an overloaded `getLegislators` function which takes, for example, a congressional district.

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
kevincianfarini: What would graceful look like in this scenario? If we aren't able to provide any legislators to the changemaker using the tool, then the remainder of the flow is useless. I think the best we can do is 500 and log the result. 

I plan on following this PR up at some point with a global exception handler and more robust logging.

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
kevincianfarini: Can do

backend/src/main/kotlin/com/climatechangemakers/act/di/ServiceModule.kt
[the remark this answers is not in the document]
kevincianfarini: Ended up going with `@Named` because it's slightly simpler.

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
kevincianfarini: Sorry didn't see your example. I like that.

kevincianfarini:
@mcchang I'm going to merge for right now. I can do a follow up PR with any further comments.

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
mcchang: Graceful would be returning something some sort of error to the frontend, perhaps with an error code or message to display to the user.  Makes sense to follow up with what you suggested

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
mcchang: a public getter on `GeocodioLegislator` might be useful if any other function or method works with `GeocodioApiResults` and `GeocodioLegislators`

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
kevincianfarini: For sure, global exception handling is the place to do this. I'll open an issue for it so we don't lose track of it.

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
kevincianfarini: I think when the need arises we can make it public and move this definition to the appropriate file.

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
kevincianfarini: #10

PR #7, opened 2021-07-31
Introduce local replica of CCM database 

kevincianfarini:
@mcchang I might need some help here getting the prod schema replicated. 

The general flow of how I'd like to do this to by exporting a sql dump file and having the postgres container ingest that on initialization using `/docker-entrypoint-initdb.d/`. I think I'm having issues exporting the schema correctly. 

So far, I've tried replicating the schema with `pg_dump`. I was originally getting some errors saying that the `postgres` role did not exist on container initialization, so I added `--no-owner` to the dump command. 

```shell
pg_dump --no-owner -s postgres -U ccm_readonly -h <DB_URL> -p 5432 > ccm_schema.sql
```
Currently, we're getting the following error. 

```shell
2021-07-31 16:17:06.801 UTC [47] ERROR:  role "readwrite" does not exist
2021-07-31 16:17:06.801 UTC [47] STATEMENT:  GRANT ALL ON SCHEMA analysis TO readwrite;
psql:/docker-entrypoint-initdb.d/ccm_schema.sql:655: ERROR:  role "readwrite" does not exist
```
This possibly has something to do with me replicating the schema using the `ccm_readonly` user. 

Ideally, I'd like no roles to be present in the development version of the database. I don't know if that's bad practice or not, so I leave that decision up to you. 

I'm fairly unfamiliar with Postgres, so a nudge in the right direction would be helpful.


kevincianfarini:
I ended up adding `--no-acl` and I think that did the trick. I'm able to connect to an empty db locally now. 

```shell
$ pgcli -h 0.0.0.0 -p 5432 -U postgres -W 
Password for postgres: 
Server: PostgreSQL 12.5
Version: 3.1.0
Chat: https://gitter.im/dbcli/pgcli
Home: http://pgcli.com
postgres@0:postgres> \d+
+----------+--------------------+----------+----------+------------+---------------+
| Schema   | Name               | Type     | Owner    | Size       | Description   |
|----------+--------------------+----------+----------+------------+---------------|
| public   | actions_raw        | table    | postgres | 8192 bytes | <null>        |
| public   | attendance_preview | table    | postgres | 8192 bytes | <null>        |
| public   | contacts           | table    | postgres | 8192 bytes | <null>        |
| public   | hoa_events         | table    | postgres | 0 bytes    | <null>        |
| public   | hoa_events_id_seq  | sequence | postgres | 8192 bytes | <null>        |
| public   | luma_attendance    | table    | postgres | 8192 bytes | <null>        |
| public   | segments           | view     | postgres | 0 bytes    | <null>        |
+----------+--------------------+----------+----------+------------+---------------+
SELECT 7
Time: 0.046s
postgres@0:postgres>

```

kevincianfarini:
I'm going to merge this for now. Will follow up any comments with additional PRs.

README.md
```diff
@@ -14,12 +14,26 @@ Climate Changemakers is developing a tool which allows constituents to reach out
 For a local development environment to work, some API keys need to be set as environment variables. 
 
 * [Geocodio API Key](https://www.geocod.io/)
-  * `export GEOCODIO_API_KEY=<key>`
 
 
 ### Docker
 
-The backend & frontend include a `Dockerfile`. You should have no need to install JDK/JRE/NODE if you're running the application through Docker. The local development process should be frictionless with `docker-compose`.
+The backend, frontend, and database include a `Dockerfile`. You should have no need to install JDK/JRE/NODE/PostgresSQL if you're running the application through Docker. The local development process should be frictionless with `docker-compose`.
```
mcchang, line 21: `docker-compose up`? Or does it work without running `docker-compose` on its own bring up everything now?

mcchang (commented):
Seems fine for now if you got it working.  If we want to clean this up we can probably tweak the DDL https://github.com/climatechangemakers/airtable_sync/commit/257779596a4975f8db2bed55764bb9dd1acb3081 somehow by specifying and creating a `ccm-act` user or something like that.

README.md
kevincianfarini: slightly below this after explaining what's necessary in the `.env` file we instruct the user to do `docker-compose up -d`. Everything else happens behind the scenes.

kevincianfarini:
@mcchang my end goal is to export a mix of schema and data once we get the LCV scores in the prod db. I need to figure out how that's done, but I don't know if that's something we can achieve with the airtable DDL. Thoughts?

README.md
mcchang: oh, I misinterpreted this section then.  Makes sense

mcchang:
> @mcchang my end goal is to export a mix of schema and data once we get the LCV scores in the prod db. I need to figure out how that's done, but I don't know if that's something we can achieve with the airtable DDL. Thoughts?

For now I would keep it simple and have the 2 lcv score data csvs in the repo and one `bootstrap_db` script and 
1. Runs the ddl
2. runs the postgres copy command to load the data from the csvs into the proper tables

Down the line we might want to use a proper ORM that generates and runs migrations for us.

kevincianfarini:
@mcchang 

> Down the line we might want to use a proper ORM that generates and runs migrations for us.

If we want, we can allow SQLDelight to handle migrations for us. I have that explicitly disabled since this action tool won't be the sole owner of the database. Right now we just describe what the current schema looks like and SQLDelight generates Kotlin mappings and plenty of other fancy things for us. 

> Runs the ddl

I toyed with this idea too. If we need to, we can fall back to this. I'd really prefer to have a SQL fixture which reflects the prod db as closely as possible and I also think it might be easier long term. It just depends if I can coerce `pg_dumb` to dumb strictly schema for some tables and while dumping schema + data for the lcv scores.

PR #11, opened 2021-08-06
Query real LCV data from the database 

### LINKED ISSUE #5: Return real LCV scores for representatives

kevincianfarini:
Prior to this commit, LCV data was faked for the purpose of
  fleshing out an API a consumer could adhere to.
  
  This commit queries for LCV lifetime and yearly scores by:
  
  * Adding the SQLDelight dependency.
  * Specifying our schema for the two LCV score tables in `.sq` files.
  * Updating the `dumb_db.sh` script to replicate data from the LCV
    tables, but only schema elsewhere.
  * Adding testing which spins up ephemeral postgres docker containers
    that exist only for the duration of a test.

Closes #5


backend/build.gradle.kts
```diff
@@ -28,6 +30,14 @@ application {
     mainClass.set("com.climatechangemakers.act.ApplicationKt")
 }
 
+sqldelight {
+    database("Database") {
+        packageName = "com.climatechangemakers.act.db"
+        dialect = "postgresql"
+        deriveSchemaFromMigrations = false
```
kevincianfarini, line 37: @mcchang this is the portion that specifies if we allow SQLDelight to control DB migrations or not. Two outstanding questions that I need to clarify. 

* Is it possible for SQLDelight to own only a portion of the schema for a given database? 
* Can migrations be performed only on that portion? 

If the answers to both of those questions are yes, I think we should allow the backend to manage migrations of the tables it owns.

kevincianfarini: The above can be a different commit though. This is non-blocking.

backend/build.gradle.kts
```diff
@@ -60,4 +73,5 @@ dependencies {
 
     testImplementation("io.ktor:ktor-server-tests:$ktor_version")
     testImplementation("org.jetbrains.kotlin:kotlin-test:$kotlin_version")
+    testImplementation("org.testcontainers:postgresql:1.15.3")
```
kevincianfarini, line 76: This is a neat tool which will spin up a new `postgres` docker container instance for every unit test we have. This is useful since we're not using something like SQLite which can run as an in memory DB.

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
```diff
@@ -28,7 +28,7 @@ class LegislatorsManager @Inject constructor(
 
     geocodioLegislators.map { geocodioLegislator ->
       async {
-        val lcvScores = lcvScoreManager.getScores(geocodioLegislator.fullName).also { scores ->
+        val lcvScores = lcvScoreManager.getScores(geocodioLegislator.references.bioguide).also { scores ->
           // ensure we at least have a lifetime score for this individual
           checkNotNull(scores.firstOrNull { it.scoreType == LcvScoreType.LifetimeScore })
```
kevincianfarini, line 33: Remove this. This errors internally in `RealLCVScoreManager` now that we're using real queries.

backend/src/test/kotlin/com/climatechangemakers/act/feature/lcvscore/manager/RealLcvScoreManagerTest.kt
```diff
@@ -0,0 +1,61 @@
...
+import java.sql.DriverManager
+import kotlin.test.Test
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
+import kotlin.test.assertEquals
+
+class RealLcvScoreManagerTest {
+
+  private val connection = DriverManager.getConnection("jdbc:tc:postgresql:12.5:///my_db")
```
kevincianfarini, line 18: This special JDBC connection URL is what instructs `testcontainers` to spin up a new postgres instance.

database/dump_db.sh
```diff
@@ -1,3 +1,14 @@
 #!/bin/bash
 
-pg_dump --no-acl --no-owner -s postgres -U ccm_readonly -h ccm-db.c51ekbqkhdej.us-west-2.rds.amazonaws.com -p 5432 > ccm_schema.sql
+pg_dump --no-acl\
+       	--no-owner\
+       	--exclude-table-data "actions_raw"\
+       	--exclude-table-data "*attendanc*"\
+       	--exclude-table-data "hoa_*"\
+       	--exclude-table-data "segments"\
+       	--exclude-table-data "contacts"\
```
kevincianfarini, line 9: @mcchang grabbing schema for some tables and schema + data for others was fairly straightforward with `--exclude-table-data`. It would have been nice for something like `--include-table-data` to exist though.

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
kevincianfarini: Jk. The contract of `LCVScoreManager` should return an empty list.

kevincianfarini:
Merging for now.

backend/src/main/kotlin/com/climatechangemakers/act/feature/lcvscore/manager/RealLcvScoreManager.kt
```diff
@@ -1,18 +1,39 @@
...
 import com.climatechangemakers.act.feature.lcvscore.model.LcvScore
 import com.climatechangemakers.act.feature.lcvscore.model.LcvScoreType
+import kotlinx.coroutines.async
+import kotlinx.coroutines.withContext
 import javax.inject.Inject
+import javax.inject.Named
+import kotlin.coroutines.CoroutineContext
 
-class RealLcvScoreManager @Inject constructor() : LcvScoreManager {
+class RealLcvScoreManager @Inject constructor(
```
mcchang, line 12: is it convention to name implementations of an interface "Real$Interface"? Don't have a great name for this implementation, but even `DBLcvScoreManager` would be more descriptive.

backend/src/main/sqldelight/com.climatechangemakers.act.db/LcvLifetimeScore.sq
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE lcv_score_lifetime (
+  bioguide_id VARCHAR PRIMARY KEY,
+  score INTEGER AS Int NOT NULL
+);
+
+selectById:
+SELECT score
+FROM lcv_score_lifetime
+WHERE bioguide_id = :bioguideId
+LIMIT 1;
```
mcchang, line 10: feels like this line shouldn't be necessary if bioguide_id is the pk?

backend/src/test/kotlin/com/climatechangemakers/act/feature/lcvscore/manager/RealLcvScoreManagerTest.kt
mcchang: perhaps leave this github comment as a comment in the actual code?

backend/src/main/kotlin/com/climatechangemakers/act/feature/lcvscore/manager/LcvScoreManager.kt
```diff
@@ -4,5 +4,9 @@ import com.climatechangemakers.act.feature.lcvscore.model.LcvScore
 
 fun interface LcvScoreManager {
 
-  suspend fun getScores(legislatorName: String): List<LcvScore>
+  /**
+   * Retrieve a list of [LcvScore] for a given representative's [bioguideId].
+   * If no scores are found for this [bioguideId], an empty list is returned.
+   */
+  suspend fun getScores(bioguideId: String): List<LcvScore>
```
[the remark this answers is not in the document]
mcchang: Related, but the `LcvScore` class trying to encapsulate both yearly and lifetime scores feels strange to me.  Every time I have an instance of `LcvScore` I need to check the `scoreType` to figure out what type of data is contained in it.

Not familiar with the best way to handle this in Kotlin, but one thing we could do is to have subclasses `LcvLifetimeScore` and `LcvYearlyScore` that inherit from the base `LcvScore`.

backend/build.gradle.kts
mcchang: > Is it possible for SQLDelight to own only a portion of the schema for a given database?

I think so, if we designate a new database user with its own permissions.

> Can migrations be performed only on that portion?
That's probably a SQLDelight question, but I would assume so

backend/src/main/kotlin/com/climatechangemakers/act/feature/lcvscore/manager/RealLcvScoreManager.kt
kevincianfarini: `Real` is a pattern we use a Square. I like the name `DatabaseLcvScoreManager`, though. I can follow up with that.

backend/src/main/kotlin/com/climatechangemakers/act/feature/lcvscore/manager/LcvScoreManager.kt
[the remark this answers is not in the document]
kevincianfarini: I like splitting up the interface. 

However, doesn't feel like a good place to introduce a class hierarchy. 

The way this is defined uses a [sum type](https://en.wikipedia.org/wiki/Tagged_union) to describe the shape of the data. In Kotlin, we can use exhaustive `when` statements over sum types which prevents a whole class of bugs. 

Kotlin `sealed` interfaces and classes are very similar to swift enum classes with associated data.

PR #12, opened 2021-08-13
Address remarks in #11

kevincianfarini:


backend/build.gradle.kts
```diff
@@ -32,9 +32,10 @@ application {
 
 sqldelight {
     database("Database") {
-        packageName = "com.climatechangemakers.act.db"
+        packageName = "com.climatechangemakers.act.database"
```
kevincianfarini, line 35: This is necessary because of a bug in SQLDelight that I just reported. See https://github.com/cashapp/sqldelight/issues/2381

kevincianfarini:
I'm going to eagerly merge since this is a follow up to #11

mcchang:
looks good!

PR #13, opened 2021-08-17
Allow backend and database container communication

kevincianfarini:


kevincianfarini:
I'm going to eagerly merge this because the current `docker-compose` setup is broken. This fixes it.

PR #17, opened 2021-08-19
Jnugent/connect to backend

jacknugent:
- Add environment variables for all ports and use them in docker-compose
- Add proxy configuration in package.json
- Add initiate-action API consumption (just console.log the results, will use results in future PR)
- Rename `Zip` and `Address` to align with backend variable names (`postalCode` and `streetAddress`)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,11 +4,8 @@ services:
     frontend:
       build: ./frontend
       image: frontend:latest
-      environment:
-        PORT: 3000
-        PROXY_API: http://backend:8080/
       ports:
-        - 3000:3000
+        - ${REACT_APP_PORT}:${REACT_APP_PORT}
     backend:
       build: ./backend
       image: backend:latest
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -29,4 +26,4 @@ services:
         - POSTGRES_USER=${POSTGRES_USER}
         - POSTGRES_DB=${POSTGRES_DB} 
       ports:
-        - 5432:5432
+        - ${POSTGRES_PORT}:${POSTGRES_PORT}
```
kevincianfarini, line 29: Love this, I knew this was something I was missing.

frontend/package.json
```diff
@@ -24,6 +24,7 @@
       "react-app"
     ]
   },
+  "proxy": "http://backend:8080",
```
kevincianfarini, line 27: Is there any way to make this depend on an environment variable? How would this work if this were deployed to production?

kevincianfarini (approved):

frontend/package.json
jacknugent: In production, this would work fine if the backend is on port 8080.

We could use [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware) to make this depend on an environment variable, as the [create react docs](https://create-react-app.dev/docs/proxying-api-requests-in-development/) suggest, but there seems to be a bug in that library. I created an [issue](https://github.com/climatechangemakers/climate-changemakers-act/issues/18) elaborating.

Issue #25, opened 2021-08-20
Add landing page privacy controls 


kevincianfarini:
We require two privacy controls to be displayed before a changemaker pursues action. 

* I consent to allow Climate Changemakers to store my email address and the policymakers I contact to track our collective impact.
  
  * Consent required  
* Yes, I would like to receive occasional informational emails from Climate Changemakers!

  * Consent optional 

This change will require an API change as well as front end work.

PR #26, opened 2021-08-20
Include privacy controls in API request

kevincianfarini:
References #25


kevincianfarini:
@jacknugent this PR broke the current state of the front end. We require these values to be present in the request. 

The good news: The error handling looks good!

PR #28, opened 2021-08-20
Implement global exception handler

### LINKED ISSUE #10: Implement Global Exception handler

kevincianfarini:
Closes #10

### DIFF
--- a/backend/src/main/kotlin/com/climatechangemakers/act/plugins/ExceptionHandler.kt
+++ b/backend/src/main/kotlin/com/climatechangemakers/act/plugins/ExceptionHandler.kt
@@ -1,0 +1,24 @@
+package com.climatechangemakers.act.plugins
+
+import io.ktor.application.Application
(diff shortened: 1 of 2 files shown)


backend/src/main/kotlin/com/climatechangemakers/act/plugins/ExceptionHandler.kt
```diff
@@ -0,0 +1,24 @@
...
+  install(StatusPages) {
+
+    exception<SerializationException> { cause ->
+      call.respond(HttpStatusCode.BadRequest, cause.message ?: "")
+    }
+
+    exception<HttpException> { cause ->
+      // TODO(kcianfarini) log this
+      call.respond(HttpStatusCode.InternalServerError, cause.message())
+    }
```
mcchang, line 22: not familiar with ktor, but should we have some sort of generic exception handler?

PR #32, opened 2021-08-21
Fix SQLDelight file paths

kevincianfarini:
Prior to this commit, the SQLDelight `.sq` file path was a single
folder separated by `.`. This was incorrect and actually exposed a bug
in SQLDelight.

That bug is fixed here
https://github.com/cashapp/sqldelight/commit/bf66bc688c5f7092c62c06d1983adf9a84aa31bc.

This commit corrects this by nesting the `.sq` files in a proper file
tree.


kevincianfarini:
eagerly merging as this impacts nobody.

PR #52, opened 2021-09-03
Add select issues page

jacknugent:
Desktop (highlighting issue 5):
image

Mobile:
image


kevincianfarini:
Looks pretty good. Is it possible to have that first row's items expand to fill the entire height of the tallest element in a row? It looks a bit strange for the issues to be misaligned like so. 

image

jacknugent:
> Looks pretty good. Is it possible to have that first row's items expand to fill the entire height of the tallest element in a row? It looks a bit strange for the issues to be misaligned like so.
> 
> image

Made that change, agree that it looks better:
image

kevincianfarini:
I think this looks much better now. Thanks! :smile:

kevincianfarini (approved):

PR #55, opened 2021-09-08
Address issue feedback from 9/7/21 meeting

jacknugent:
- Allow re-selecting an issue
- Center email input if no talking points:

image

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -24,7 +24,7 @@
       "react-app"
     ]
   },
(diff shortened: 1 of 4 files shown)


kevincianfarini:
If there's no talking points, is it possible to make the email field stretch the width of the content container?

jacknugent:
> If there's no talking points, is it possible to make the email field stretch the width of the content container?

Yes, done

kevincianfarini (approved):

PR #68, opened 2021-09-19
Get api api endpoint from env

mcchang:


frontend/src/api/ClimateChangemakersAPI.ts
```diff
@@ -1,6 +1,10 @@
 import { ActionInfo } from "../models/ActionInfo";
 import { IssuesResponse } from "../models/IssuesResponse";
 
+// const apiEndpoint = "https://changemakers-act-backend.onrender.com/"
+const apiEndpoint = process.env.REACT_APP_API_ENDPOINT || '';
```
kevincianfarini, line 5: How does this impact the anticipated bundling of the frontend into static assets which will be served from somewhere? We're not going to have an environment at that point, and the `apiEndpoint` is going to implicitly be where this content was served from, right?

frontend/src/api/ClimateChangemakersAPI.ts
```diff
@@ -1,6 +1,10 @@
...
+console.log(process.env.NODE_ENV);
```
jacknugent, line 6: Dev code?

frontend/src/api/ClimateChangemakersAPI.ts
mcchang: Yeah, ignore this PR.  It don't think it's needed.  I was just experimenting with various ways to deploy the frontend and the backend.

frontend/src/api/ClimateChangemakersAPI.ts
mcchang: yeah, this PR was a WIP.  Disregard!

PR #69, opened 2021-09-19
Include closest legislator district office phone number

### LINKED ISSUE #66: Add "Who to Call" data to `initiate-action` endpoint
Screenshot from [most recent design](https://www.figma.com/file/PEEkfIkziGfLogDHxSex9K/CC-%2F%2F-Action-Tool?node-id=205%3A2):
image

Looks like the front-end will need the following info for each representative:
- Who to call
- DC Office number
- Local Office number

Would make sense to add this onto `initiate-action` endpoint. @kevincianfarini thoughts? Not sure where this data lives, or if there can be more than 2 numbers to call for each representative.

kevincianfarini:
Closes #66


backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/model/Legislator.kt
```diff
@@ -7,7 +7,7 @@ import kotlinx.serialization.Serializable
   val name: String,
   val role: LegislatorRole,
   val siteUrl: String,
-  val phone: String,
+  val phoneNumbers: List<String>,
```
kevincianfarini, line 10: @jacknugent this is API breaking :/. Sorry about that.

frontend/src/models/ActionInfo.ts
```diff
@@ -5,6 +5,7 @@ export type ActionInfo = {
         area: {
             state: string;
             districtNumber?: number;
+            districtPhoneNumber?: string;
```
kevincianfarini, line 8: This is not something that's included in the API. @jacknugent

jacknugent: Sorry didn't read your back-end code correctly, removed

kevincianfarini: All good, thanks.

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/model/Legislator.kt
jacknugent: Adjusted the front-end type in [this commit](https://github.com/climatechangemakers/climate-changemakers-act/pull/69/commits/c845c732166f2779f351162864ca2325f02bd5dc)

jacknugent (approved):

PR #70, opened 2021-09-19
Remove LCV score existence check

kevincianfarini:
It's possible that a member of congress doesn't have a LCV score. This
could be because the MoC is new and hasn't voted on any climate
legislation yet.

### DIFF
--- a/backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
+++ b/backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
@@ -37,7 +37,7 @@ class LegislatorsManager @Inject constructor(
         geocodioLegislator.toDomainLegislator(
           state = request.state,
           districtNumber = congressionalDistrict.districtNumber,
-          lcvScores = lcvScores.await().also { check(it.isNotEmpty()) },
+          lcvScores = lcvScores.await(),
           districtPhoneNumber = districtPhoneNumber.await(),
         )
       }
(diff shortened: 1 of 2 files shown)


backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
```diff
@@ -43,7 +43,7 @@ class LegislatorsManager @Inject constructor(
         geocodioLegislator.toDomainLegislator(
           state = request.state,
           districtNumber = congressionalDistrict.districtNumber,
-          lcvScores = lcvScores.also { check(it.isNotEmpty()) }
```
kevincianfarini, line 46: @jacknugent does this break any assumptions on the frontend?

jacknugent: I added a [commit to this PR](https://github.com/climatechangemakers/climate-changemakers-act/pull/70/commits/8126d8997eecfc31b0885a59a3fbefef40d456f7) to account for optional LCV score

jacknugent (approved):

backend/src/main/kotlin/com/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
kevincianfarini: Looks good to me!

EvanHahn (approved):
I don't have much context here, but this looks reasonable to me.

PR #73, opened 2021-09-22
Rename backend package qualifier

kevincianfarini:
Climate Changemakers is not a commercial entity, and thus does not use
the `com` TLD.

This commit renamed the backend package from
`com.climatechangemakers.act` to `org.climatechangemakers.act`.


kevincianfarini:
This is a simple name change. I'm going to merge for now to avoid massive merge conflicts in the future. If y'all have comments, I can address them in a follow up.

Issue #36, opened 2021-08-21
Send legislator emails via ActionNetwork


kevincianfarini:
Relevant documentation: https://actionnetwork.org/docs/v2/

kevincianfarini:
Might be worth avoiding action network altogether and use https://www.house.gov/doing-business-with-the-house/communicating-with-congress-cwc

kevincianfarini:
There's an additional senate counterpart to the above house API. 

https://www.senate.gov/senators/scwc.htm

We should almost certainly be using those over Action Network's API.

kevincianfarini:
We got access to SCWC, waiting on access to (House) CWC. We're going to ditch Action Network. It's too complicated and does too much for our use cases. Renaming this issue. 

One interesting piece of documentation from CWC is their planned maintenance twice weekly. 

> Heads Up We perform weekly maintenance from 12 a.m. to 6 a.m. every Sunday and 5 a.m. to 7 a.m. every Wednesday. SCWC may be intermittently out during those times.

Maybe this warrants some sort of queuing service? That could be an improvement for the future later down the line. We don't need to worry about it for 1.0.

mcchang:
Would be nice to link to the xml spec for the CWC api (I couldn't easily find it via Google?)

mcchang:
> We got access to SCWC, waiting on access to (House) CWC. We're going to ditch Action Network. It's too complicated and does too much for our use cases. Renaming this issue.
> 
> One interesting piece of documentation from CWC is their planned maintenance twice weekly.
> 
> > Heads Up We perform weekly maintenance from 12 a.m. to 6 a.m. every Sunday and 5 a.m. to 7 a.m. every Wednesday. SCWC may be intermittently out during those times.
> 
> Maybe this warrants some sort of queuing service? (cc @mcchang) That could be an improvement for the future later down the line. We don't need to worry about it for 1.0.

Makes sense! We'll probably want to retry on transient errors like these.

kevincianfarini:
@mcchang I can chat with you about it offline. I don't believe the documentation is accessible by the public.

PR #76, opened 2021-09-25
Include initial serialization schema for CWC

kevincianfarini:
We don't currently have CWC API keys, but the documentation has allowed
me to get this far. Tests are actually written for serialization because
the CWC API ingests XML, which is a loosy goosey format.

References #36

### DIFF
--- a/backend/src/main/kotlin/org/climatechangemakers/act/common/model/State.kt
+++ b/backend/src/main/kotlin/org/climatechangemakers/act/common/model/State.kt
@@ -5,14 +5,15 @@ import org.climatechangemakers.act.common.serializers.StringEnum
 import org.climatechangemakers.act.common.serializers.StringEnumSerializer
 
 @Serializable(with = StateSerializer::class) enum class State(override val value: String) : StringEnum {
-  Alabama("AL"),
   Alaska("AK"),
+  Alabama("AL"),
   AmericanSamoa("AS"),
   Arizona("AZ"),
   Arkansas("AR"),
   California("CA"),
   Colorado("CO"),
   Connecticut("CT"),
+  DistrictOfColumbia("DC"),
   Delaware("DE"),
   Florida("FL"),
   Georgia("GA"),
@@ -41,6 +42,7 @@ import org.climatechangemakers.act.common.serializers.StringEnumSerializer
   NewYork("NY"),
   NorthCarolina("NC"),
   NorthDakota("ND"),
+  NorthernMarianaIslands("MP"),
   Ohio("OH"),
   Oklahoma("OK"),
   Oregon("OR"),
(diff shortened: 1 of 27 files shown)


backend/src/main/kotlin/org/climatechangemakers/act/common/serializers/DateSerializer.kt
```diff
@@ -0,0 +1,17 @@
...
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import java.text.SimpleDateFormat
+import java.util.Date
+
+object DateSerializer : KSerializer<Date> {
+  override val descriptor = PrimitiveSerialDescriptor("Date", PrimitiveKind.STRING)
+  override fun serialize(encoder: Encoder, value: Date) = encoder.encodeString(value.toEightCharFormat())
+  override fun deserialize(decoder: Decoder): Date = TODO()
```
mcchang, line 14: lol `TODO()`, I love it

backend/src/main/kotlin/org/climatechangemakers/act/common/model/State.kt
```diff
@@ -0,0 +1,68 @@
+package org.climatechangemakers.act.common.model
+
+import kotlinx.serialization.Serializable
+import org.climatechangemakers.act.common.serializers.StringEnum
+import org.climatechangemakers.act.common.serializers.StringEnumSerializer
+
+@Serializable(with = StateSerializer::class) enum class State(override val value: String) : StringEnum {
+  Alaska("AK"),
+  Alabama("AL"),
+  AmericanSamoa("AS"),
```
mcchang, line 10: Would it make sense to split out areas/territories like American Samoa and DC from this list? I think I would be surprised to see those areas as enum values.  I would just expect the 50 states.  Alternatively, we could name this something more descriptive like "JurisdictionCode" or something

mcchang (approved):
Generally looks good! Sucks that it's a XML API :(.  Wish there was something off the shelf we could use to communicate with CWC

backend/src/main/kotlin/org/climatechangemakers/act/common/serializers/DateSerializer.kt
kevincianfarini: It's shorthand for `throw NotImplementedError`!

backend/src/main/kotlin/org/climatechangemakers/act/common/model/State.kt
kevincianfarini: I think naming is the issue here. I thought of the same thing.

This isn't a comprehensive list of all US States + Territories. Instead, it's a list of all places the US has representation for in congress. 

I'm thinking we all it `RepresentedArea`.

PR #79, opened 2021-09-27
Replace geocodio congressional data with `member_of_congress` table

### LINKED ISSUE #38: Need State Code/District/Twitter Handle on /initiate-action endpoint
`/initiate-action` needs the following info:
- 2-digit state code (I could pull that from the form, but I would prefer not to)
- District 
- Twitter handle of congressperson

kevincianfarini:
Closes #38.


jacknugent (approved):

kevincianfarini:
@mcchang this isn't a blocker for this PR, but I noticed that `render` doesn't seem to update with the PR? This PR introduces an additional `twitter` field to the API contract, and I'm not seeing it. I suspect that render is either misconfigured to look at the wrong branch or caching docker images. Can you look into this?

PR #86, opened 2021-10-01
Complete async work before sending http response

### LINKED ISSUE #84: PSQLExceptions are silenced

kevincianfarini:
Closes #84

### DIFF
--- a/backend/src/main/kotlin/org/climatechangemakers/act/common/extension/ApplicationCallExtensions.kt
+++ b/backend/src/main/kotlin/org/climatechangemakers/act/common/extension/ApplicationCallExtensions.kt
@@ -1,0 +1,7 @@
+package org.climatechangemakers.act.common.extension
+
+import io.ktor.application.ApplicationCall
+import io.ktor.http.HttpStatusCode
+import io.ktor.response.respond
+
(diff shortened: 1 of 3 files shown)


kevincianfarini:
It was possible for us to send an http response before all of the async work it was handling was complete. This allowed exceptions that were bubbled up from the async work to be dropped on the ground without the exception handler responding appropriately. 

Consider the following example: 

```kotlin
suspend fun myHttpRequest(call: ApplicationCall) = coroutineScope {
  val request = call.recieve()
  launch { ... } // async method builder returns immediately, but errors in the background
  call.resond(...)
}
```
Because the call to `launch` does work in the background, and the calling method can keep on running, `call.respond` gets invoked before an exception has a chance to get bubbled up. The solution to this is to ensure that all async work is completed before `call.respond` is invoked. `coroutineScope` handles this for us. 

```kotlin
suspend fun myHttpRequest(call: ApplicationCall) {
  val request = call.recieve()
  
  coroutineScope { // suspend until all inner async work is completed 
    launch { ... }  // errors
    launch { ... }
  }

  call.resond(...)
}
```

When the call to `launch` errors, it bubbles up through `coroutineScope` and `myHttpRequest` all the way up to the built in exception handler, and fails the request.

kevincianfarini:
I'm going to eagerly merge for now. Will follow resolve possible review comments in another PR. just @ me.

PR #88, opened 2021-10-02
Add library-of-congress-topics route

### LINKED ISSUE #87: Provide an endpoint to GET the library of congress topics

kevincianfarini:
Closes #87


backend/src/main/kotlin/org/climatechangemakers/act/feature/values/controller/ValuesController.kt
```diff
@@ -25,6 +28,14 @@ class ValuesController @Inject constructor(private val json: Json) {
       status = HttpStatusCode.OK,
     )
   }
+
+  suspend fun libraryOfCongressTopicValues(call: ApplicationCall) {
+    call.respondText(
```
kevincianfarini, line 33: unfortunately Ktor can't pick up on this serialization schema automatically because for some reason, doing `@Serializable(with = TopicSerializer::class)` isn't resolved at compile time. I think this is s a shortcoming of the kotlinx.serialization library.

PR #102, opened 2021-10-05
Add necessary fields to `SendEmailRequest`

### LINKED ISSUE #56: Stabilize `send-email` API contract
I'm thinking the endpoint could look something like the following. 

```typescript
POST /api/send-email

{
  "relatedIssueId": int, // the issue a changemaker is writing their legislators about
  "emailBody": string,
  "contactedBioguideIds": string[] // the list of legislators the email is being sent to
}
```

kevincianfarini:
This commit adds fields necessary to make a (S)CWC contact http call.
Some of this information has already been collected from the user as
part of finding their members of congress -- we should reuse that
information versus redundantly asking for it again.

The `relatedTopics` and `contactedBioguideIds` always expect to have at
least one element in them.

Closes #56

### DIFF
--- a/backend/src/main/kotlin/org/climatechangemakers/act/feature/action/model/SendEmailRequest.kt
+++ b/backend/src/main/kotlin/org/climatechangemakers/act/feature/action/model/SendEmailRequest.kt
@@ -23,5 +23,8 @@ import org.climatechangemakers.act.feature.communicatewithcongress.model.Topic
   init {
     require(contactedBioguideIds.isNotEmpty())
     require(relatedTopics.isNotEmpty())
+    require(contactedBioguideIds.size <= 3)
+    require(firstName.isNotBlank())
+    require(lastName.isNotBlank())
   }
 }
--- a/backend/src/main/kotlin/org/climatechangemakers/act/feature/action/model/SendEmailRequest.kt
+++ b/backend/src/main/kotlin/org/climatechangemakers/act/feature/action/model/SendEmailRequest.kt
@@ -1,10 +1,27 @@
 package org.climatechangemakers.act.feature.action.model
 
 import kotlinx.serialization.Serializable
+import org.climatechangemakers.act.common.model.RepresentedArea
+import org.climatechangemakers.act.feature.communicatewithcongress.model.Prefix
+import org.climatechangemakers.act.feature.communicatewithcongress.model.Topic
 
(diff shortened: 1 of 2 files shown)


backend/src/main/kotlin/org/climatechangemakers/act/feature/action/model/SendEmailRequest.kt
```diff
@@ -1,10 +1,27 @@
...
+  val postalCode: String,
+  val relatedTopics: List<Topic>,
+  val emailSubject: String,
   val emailBody: String,
+  val relatedIssueId: Long,
   val contactedBioguideIds: List<String>,
-)
\ No newline at end of file
+) {
+  init {
```
EvanHahn, line 23: Do we want to add additional validations? For example, `firstName` probably has to be nonempty.

Also, is there any security risk with any of these? I could imagine someone knocking the server over with a giant array or a long name.

backend/src/main/kotlin/org/climatechangemakers/act/feature/action/model/SendEmailRequest.kt
```diff
@@ -1,10 +1,27 @@
...
   val originatingEmailAddress: String,
-  val relatedIssueId: Long,
+  val title: Prefix,
+  val firstName: String,
+  val lastName: String,
+  val streetAddress: String,
+  val city: String,
+  val state: RepresentedArea,
+  val postalCode: String,
+  val relatedTopics: List<Topic>,
```
EvanHahn, line 17: Is this resilient given that representatives might change the topic strings on a whim? (Maybe this comment is out of scope for this PR.)

EvanHahn (approved):
LGTM. Left some comments that shouldn't block a merge.

backend/src/main/kotlin/org/climatechangemakers/act/feature/action/model/SendEmailRequest.kt
kevincianfarini: These topics are sanctioned by the CWC api!

backend/src/main/kotlin/org/climatechangemakers/act/feature/action/model/SendEmailRequest.kt
kevincianfarini: Great point!

Issue #100, opened 2021-10-05
Implement Social Media section 


jacknugent:
From the designs:
<img width="1060" alt="image" src="https://user-images.githubusercontent.com/31449177/135947433-538d8c8b-8e5e-4478-89be-d26de243f957.png">

The textbox should be read-only, and the "Send Tweet" button should send the user to twitter.com with the textbox message pre-populated. Alternatively, the user can click "Skip" to go to the next step.

Currently, there is an existing `PostOnSocial` Component in React that needs to be expanded to include these designs.

Where we source our icons (such as the megaphone next to the title) is TBD

kevincianfarini:
The endpoint to record a tweet action is `/api/log-tweet`. When a user clicks "send tweet" a new tab should be opened with a pre composed tweet, and should simultaneously make a call to log their action.

EvanHahn:
I'll plan to look into this one. A few questions, in no particular order:

1. Should anything visibly change when you click Send Tweet? I assume it should take you to the next step so it's there when you return, but maybe that's wrong.
2. Can you provide an example payload to send to the API? I assume it's a `POST` or `PUT`, but should any additional data be sent? (I can probably paw through the code to answer this question myself.)
3. Those designs feature highlighted hashtags and @-mentions. That's totally doable but is a little more challenging than a simple `<textarea>`. Would it be okay to omit that in a first version?
4. What retry logic, if any, should the API request have? It seems like it should be fire-and-forget and it's okay if there's an error, but maybe we want to retry it?

I haven't looked at the existing code yet, so maybe a lot of these questions have obvious answers.

EvanHahn-Signal:
> Those designs feature highlighted hashtags and @-mentions. That's totally doable but is a little more challenging than a simple `<textarea>`. Would it be okay to omit that in a first version?

For posterity, we had a call and agreed to just do a simple `<textarea>` for now.

EvanHahn:
> Those designs feature highlighted hashtags and @-mentions. That's totally doable but is a little more challenging than a simple `<textarea>`. Would it be okay to omit that in a first version?

For posterity, we had a call and agreed to just do a simple `<textarea>` for now.

kevincianfarini:
I just came to the realization that the pre composed tweet should be something driven by the server. I need to add an endpoint for it. 

Created #105

PR #110, opened 2021-10-08
Add "Post on Social" section

EvanHahn:
Screenshot

This pull request implements a basic "Post on Social" section as described in #100.

Some notable technical pieces:

- The backend now returns a JSON object for the pre-composed tweet instead of plain text. This is more extensible and simplifies the frontend code, but I can revert it if needed.
- To enforce tweet validity, I pulled in Twitter's official [tweet-text library](https://github.com/twitter/twitter-text). This is more accurate than a simpler check like `tweet.length > 0 && tweet.length <= 280`, but [it's a nontrivial module](https://bundlephobia.com/package/twitter-text@3.1.0) which we might want to skip for now.
- The pre-composed tweet is loaded as soon as `<TakeActionPage>` mounts so users will probably never see it mid-load, but there's some code to handle that just in case.
- `scrollIntoView` doesn't work if the page is in the background, at least in my testing. This happens when we open a new window to post a tweet. I did a bunch of gymnastics to make this work. I don't love what I came up with, so suggestions are welcome.

A few other things:

- The copy and stylings seemed TBD so I didn't implement this exactly as the designs showed.
- I tried my best to keep the code style consistent, but I fear that I let a little too much of my strange code style through. Please let me know if I deviated too far. Also, I'd love to use an auto-formatter like [Prettier](https://prettier.io/) later on, but that's out of scope for this change.


frontend/src/Pages/TakeAction/TakeActionPage.tsx
```diff
@@ -18,13 +20,86 @@ export default function TakeActionPage() {
...
+
+    const selectedIssueId = selectedIssue?.id;
+    useEffect(() => {
+        if (!selectedIssueId) {
+            return;
+        }
+
+        setPreComposedTweet({ status: "loading" });
+
+        let isCanceled = false;
```
EvanHahn, line 81: The API functions don't support cancellation, so this is a simpler way to clean up the effect.

frontend/src/common/lib/twitter.ts
```diff
@@ -0,0 +1,5 @@
+export function getPostTweetUrl(text: string): string {
+    const url = new URL("https://twitter.com/intent/tweet");
```
EvanHahn, line 2: `URL` [doesn't work on all browsers](https://caniuse.com/url). Just want to call that out.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+            }
+
+            const onClickSendTweet = once(() => {
+                setHasClickedLink(true);
+
+                logTweet();
+
+                // We don't want to update the state until the tweet page has opened.
+                // We either wait for the page to be backgrounded or 1 second, whichever comes first.
+                const done = () => {
```
jacknugent, line 58: Instead of comment, give a more meaningful name, like `updateStateAfterTweetPageOpened`. In that case, would make more sense to move `if (document.hidden)` into this function

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+                <Form as="div">
+                    <Form.Group>
+                        <Form.Label htmlFor="draft-tweet-input">
+                            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec a diam lectus!
+                        </Form.Label>
+                        <InputGroup hasValidation>
+                            <Form.Control
+                                as="textarea"
+                                rows={4}
+                                // We rely on the twitter-text package to enforce lengths, but we have this just in case of a pathologically long string.
```
jacknugent, line 86: Don't think think this comment is necessary

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+                                maxLength={1000}
+                                id="draft-tweet-input"
+                                placeholder="Compose your tweet"
+                                isInvalid={!isTweetValid}
+                                disabled={isSocialPosted}
+                                value={tweet}
+                                onChange={(event) => {
```
jacknugent, line 93: Don't need parens or `{}`:
```javascript
onChange={e => setTweet(e.target.value)}
```

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+                            </Form.Control.Feedback>
+                        </InputGroup>
+                    </Form.Group>
+                    <div className="row mt-3">
+                        <div className="col d-flex">
+                            <Button
+                                className="flex-grow-1 mr-2"
+                                variant="secondary"
+                                disabled={isSocialPosted}
+                                onClick={() => {
```
jacknugent, line 108: Don't need `{}`

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+                                )}
+                                Send Tweet
+                            </Button>
+                        </div>
+                    </div>
+                </Form>
+            );
+            break;
+        }
+        default:
```
jacknugent, line 142: This default will never happen, can remove and can remove `MissingCaseError`. Also think switch can be replaced entirely with ternary expressions and JSX embedded in main return function --- see above

backend/src/main/kotlin/org/climatechangemakers/act/feature/issue/controller/IssueListController.kt
```diff
@@ -26,6 +27,6 @@ class IssueListController @Inject constructor(
   )
 
   suspend fun respondPreComposedTweet(call: ApplicationCall, issueId: Long) = call.respond(
-    manager.getPreComposedTweetForIssue(issueId)
+    PreComposedTweetResponse(manager.getPreComposedTweetForIssue(issueId))
```
kevincianfarini, line 30: Rather than allocating the PreComposedTweetResponse in the controller, i think it should be done from the manager. Thoughts?

frontend/package.json
```diff
@@ -9,11 +9,13 @@
     "bootstrap": "^5.1.0",
     "canvas-confetti": "^1.4.0",
     "history": "^4.10.1",
+    "lodash": "^4.17.21",
     "react": "^17.0.2",
     "react-bootstrap": "^2.0.0-beta.5",
     "react-dom": "^17.0.2",
     "react-router-dom": "^5.3.0",
     "react-scripts": "4.0.3",
+    "twitter-text": "^3.1.0",
```
kevincianfarini, line 18: What are the benefits/drawbacks of having this validation live server side? Ideally, this would even be done at the database level, but i am unsure of what validations are being performed.

frontend/src/Pages/TakeAction/TakeActionPage.tsx
jacknugent: What do you mean by this?

frontend/src/Pages/TakeAction/TakeActionPage.tsx
```diff
@@ -18,13 +20,86 @@ export default function TakeActionPage() {
...
+        (async () => {
+            const response = await preComposedTweetAPI(selectedIssueId);
+            if (isCanceled) {
+                return;
+            }
+            if (response.successful && response.data) {
+                setPreComposedTweet({
```
jacknugent, line 88: Why not just have three separate state:
`preComposedTweet` <-- string? that eventually gets `response.data.tweet`
`errorMessage` <-- string? with possible error message
`isLoading` <-- bool for loading pre-composed tweet

Those states are pretty self-explanatory, I think this new type complicates things.

kevincianfarini (approved):

frontend/src/common/lib/Loadable.ts
```diff
@@ -0,0 +1,6 @@
+type Loadable<ValueT, ErrorT> =
+  | { status: "loading" }
+  | { status: "loaded"; value: ValueT }
+  | { status: "failed"; error: ErrorT };
```
jacknugent, line 4: If this is only used in twitter component, easier to just have at top of PostOnSocial component (although I think this type isn't necessary -- see comment above)

frontend/src/common/lib/twitter.test.ts
```diff
@@ -0,0 +1,11 @@
+import { getPostTweetUrl } from "./twitter";
+
+describe("Twitter utilities", () => {
```
jacknugent, line 3: Yes testing!!!

frontend/src/common/lib/twitter.ts
jacknugent: We are not supporting IE11

frontend/src/common/api/ClimateChangemakersAPI.ts
```diff
@@ -66,4 +67,11 @@ export const issueAPI = () =>
...
 export const areasAPI = () =>
-    get<AreasResponse>("/values/areas");
\ No newline at end of file
+    get<AreasResponse>("/values/areas");
+
+export const preComposedTweetAPI = (issueId: number) =>
+    get<PreComposedTweetResponse>(`/issues/${issueId}/precomposed-tweet`)
+
+export const logTweetAPI = async (originatingEmailAddress: string, relatedIssueId: number, contactedBioguideIds: ReadonlyArray<string>): Promise<void> => {
+    await post<unknown>("/log-tweet", { originatingEmailAddress, relatedIssueId, contactedBioguideIds });
```
jacknugent, line 76: Why `unknown`? What does this actually return?

frontend/package.json
```diff
@@ -9,11 +9,13 @@
     "bootstrap": "^5.1.0",
     "canvas-confetti": "^1.4.0",
     "history": "^4.10.1",
+    "lodash": "^4.17.21",
```
jacknugent, line 12: I'm not convinced `lodash` is necessary. ES6 seems to make it mostly obsolete. May potentially remove/keep after we chat on why we need `once` function from lodash

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+            break;
+        case "failed":
+            contents = <p>Failed to load. Please try refreshing the page.</p>;
+            break;
+        case "loaded": {
+            const { weightedLength, valid: isTweetValid } = twitterText.parseTweet(tweet);
+            const canSubmit = !isSocialPosted && isTweetValid;
+
+            // We always render an error message, even if we don't display it.
+            let errorMessage = "Your tweet is invalid. Is it too long?";
```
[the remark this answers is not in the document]
jacknugent: Even better, just add the ternary logic directly into the Form.Control.Feedback:
```javascript
<Form.Control.Feedback type="invalid" tooltip>
   {weightedLength === 0 ? "You need to enter a tweet."
      : weightedLength > 280 ? `Your tweet is ${weightedLength} characters but 280 is the maximum.`
         : "Your tweet is invalid. Is it too long?";}
</Form.Control.Feedback>
```

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+            if (weightedLength === 0) {
+                errorMessage = "You need to enter a tweet.";
+            } else if (weightedLength > 280) {
+                errorMessage = `Your tweet is ${weightedLength} characters but 280 is the maximum.`;
+            }
+
+            const onClickSendTweet = once(() => {
+                setHasClickedLink(true);
+
+                logTweet();
```
jacknugent, line 54: I think it would be clearer to remove this prop and just do the logTweetAPI call within this component

jacknugent (changes requested):

backend/src/main/kotlin/org/climatechangemakers/act/feature/issue/controller/IssueListController.kt
EvanHahn: Up to you. I assumed that the controller, not the manager, should know about the shape of the HTTP response body. But I don't know—lemme know if you want me to move it.

frontend/package.json
EvanHahn: I don't think we should put the validation on the server. Because we're just composing a draft, we'd have to make repeated requests to the server just to validate something Twitter will check anyway.

As an alternative, I could just do basic checks on the length and let the actual Twitter site do the rest.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
EvanHahn: The benefit is to add type checking to ensure that this case will never happen. If we update the type to allow a fourth option, it won't compile, (hopefully) saving us from a bug.

But I'll remove it, as I don't think it's warranted for such a small project.

frontend/src/common/api/ClimateChangemakersAPI.ts
EvanHahn: It currently returns a 204 with no response body, so I suppose it'd be `void`. I chose `unknown` because we won't have to keep it in sync with the server. But I'm happy to change it.

backend/src/main/kotlin/org/climatechangemakers/act/feature/issue/controller/IssueListController.kt
kevincianfarini: I'd prefer for this PR to remain consistent with the other managers. However, I do agree that the controllers should be the entities that know about request/response objects.

jacknugent:
image

image

Styling changes to make section look more like the designs:
- Make title pink (I added the “text-pink” class for this)
- Less spacing between title and description
- Add description from design
- Make buttons same color as design
- Slightly more spacing between `<hr />` and title
- Should we stack buttons on mobile? Up to you
- Add icon from Figma

frontend/src/Pages/TakeAction/TakeActionPage.tsx
EvanHahn: You're right.

More accurately, you don't need to clean up HTTP requests directly, but you _do_ probably want to clean up anything that follows the HTTP request. For example, this component has some problems:

```jsx
function MyComponent({ userId }) {
  const [userName, setUserName] = useState('');

  useEffect(() => {
    setUserName('');
    (async () => {
      setUserName(await fetchUserName(userId));
    })();
  }, [userId]);

  return <div>{userId} has name {userName}</div>;
}
```

There are two issues here.

1. If the component unmounts while a request is in flight, additional things can happen after the component is unmounted.

   In the example above, `setUserName` will be called after the component has unmounted, which is a no-op. That causes React to give the following warning:

   > Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.

   This is probably fine, but other side-effects could run later which we don't want. For example, we could kick off a second HTTP request, or update `document.title`, etc.

    I made [a small example app](https://codesandbox.io/s/goofy-panka-wfbb0) to demonstrate this problem.
2. The bigger problem, IMO, is that the UI can get messed up.

   For example, imagine that I change the `userId` prop twice in rapid succession. After the first request finished, we'd see the first user's name until the second request finished. And in a world where the _second_ request finished first, which could happen, we'd see the wrong user name in the UI once everything had settled down.

   I made [another small example app](https://codesandbox.io/s/suspicious-silence-ovxwh) to demonstrate this issue. Try changing the current user rapidly, and you'll see the wrong values in some situations.

If we don't clean up this request and the selected issue changes, we won't cancel the request and we might have the wrong pre-composed tweet.

frontend/src/Pages/TakeAction/TakeActionPage.tsx
```diff
@@ -18,13 +20,86 @@ export default function TakeActionPage() {
...
+    useEffect(() => {
+        if (!selectedIssueId) {
+            return;
+        }
+
+        setPreComposedTweet({ status: "loading" });
+
+        let isCanceled = false;
+        (async () => {
+            const response = await preComposedTweetAPI(selectedIssueId);
```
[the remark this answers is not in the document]
EvanHahn: I didn't do this because I wanted the tweet to be loaded by the time the user reached the step, and `<PostOnSocial>` isn't mounted when the page starts. I could:

1. keep things as is.
2. move the request into `<PostOnSocial>`, which would cause users to see a loading state. (Technically, they could see this state today, but they'd have to move through the steps pretty quickly, or be on a spotty internet connection.)
3. keep `<PostOnSocial>` mounted at all times, but don't show it until the step is ready

Any preference?

frontend/src/Pages/TakeAction/TakeActionPage.tsx
```diff
@@ -52,7 +127,19 @@ export default function TakeActionPage() {
                     {isPhoneCallMade &&
                         <>
                             <hr id="post_on_social" />
-                            <PostOnSocial isSocialPosted={isSocialPosted} setIsSocialPosted={setIsSocialPosted} />
+                            <PostOnSocial
+                                isSocialPosted={isSocialPosted}
+                                setIsSocialPosted={setIsSocialPosted}
+                                preComposedTweet={preComposedTweet}
+                                logTweet={() => {
+                                    const { initiatorEmail: email, legislators } = actionInfo;
```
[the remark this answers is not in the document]
EvanHahn: Done in 4bb3ac6efd7272eeb9e4f0d5afb646860f365a62.

frontend/src/Pages/TakeAction/TakeActionPage.tsx
```diff
@@ -52,7 +127,19 @@ export default function TakeActionPage() {
...
+                                    const bioguideIds = legislators.map(l => l.bioguideId);
+                                    logTweetAPI(email, selectedIssue.id, bioguideIds).catch(err => {
+                                        // It's okay if this request fails.
```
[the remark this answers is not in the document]
EvanHahn: Removed in 633dbb22db63558fea0ab322599fe17962685b32.

frontend/src/Pages/TakeAction/TakeActionPage.tsx
```diff
@@ -52,7 +127,19 @@ export default function TakeActionPage() {
...
                             <hr id="post_on_social" />
-                            <PostOnSocial isSocialPosted={isSocialPosted} setIsSocialPosted={setIsSocialPosted} />
+                            <PostOnSocial
+                                isSocialPosted={isSocialPosted}
+                                setIsSocialPosted={setIsSocialPosted}
+                                preComposedTweet={preComposedTweet}
+                                logTweet={() => {
+                                    const { initiatorEmail: email, legislators } = actionInfo;
+                                    const bioguideIds = legislators.map(l => l.bioguideId);
+                                    logTweetAPI(email, selectedIssue.id, bioguideIds).catch(err => {
```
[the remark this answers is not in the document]
EvanHahn: Done in 558cc98c3bc9a8b134de0274f84b6645aeeafb8b (didn't wind up using an IIFE).

frontend/src/Pages/TakeAction/TakeActionPage.tsx
```diff
@@ -18,13 +20,86 @@ export default function TakeActionPage() {
     const [isSocialPosted, setIsSocialPosted] = useState(false);
     const [selectedIssue] = useSessionStorage<Issue | undefined>("selectedIssue");
     const [actionInfo] = useSessionStorage<ActionInfo | undefined>("actionInfo");
+    const [preComposedTweet, setPreComposedTweet] = useState<Loadable<string, string>>({ status: "loading" });
 
-    const scrollToId = (id: string) =>
-        document.getElementById(id)?.scrollIntoView()
+    // `scrollIntoView` doesn't work when the window is backgrounded, so we have some complex logic to enqueue a scroll if the window is invisible.
```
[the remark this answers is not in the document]
EvanHahn: Simplified it with a "Done Tweeting" button, like we discussed "offline", in fae83afac3d2a36dc309111017554bba52f803ae.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+
+            // We always render an error message, even if we don't display it.
+            let errorMessage = "Your tweet is invalid. Is it too long?";
+            if (weightedLength === 0) {
+                errorMessage = "You need to enter a tweet.";
+            } else if (weightedLength > 280) {
+                errorMessage = `Your tweet is ${weightedLength} characters but 280 is the maximum.`;
+            }
+
+            const onClickSendTweet = once(() => {
```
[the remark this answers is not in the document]
EvanHahn: Done. Sidestepped this in fae83afac3d2a36dc309111017554bba52f803ae.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
EvanHahn: Done; sidestepped this in fae83afac3d2a36dc309111017554bba52f803ae.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+                    clearWait();
+                    setIsSocialPosted(true);
+                };
+
+                const onVisibilityChange = () => {
+                    if (document.hidden) {
+                        done();
+                    }
+                };
+                document.addEventListener('visibilitychange', onVisibilityChange, false);
```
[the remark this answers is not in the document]
EvanHahn: Removed in fae83afac3d2a36dc309111017554bba52f803ae.

frontend/package.json
EvanHahn: Removed it in fae83afac3d2a36dc309111017554bba52f803ae.

frontend/package.json
EvanHahn: Uninstalled the library and used a simpler check in 49a640ace8194f0573b71dc43f3e631132661890.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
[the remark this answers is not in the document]
EvanHahn: Done in 49a640ace8194f0573b71dc43f3e631132661890.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+            contents = <p>Loading...</p>;
+            break;
+        case "failed":
+            contents = <p>Failed to load. Please try refreshing the page.</p>;
+            break;
+        case "loaded": {
+            const { weightedLength, valid: isTweetValid } = twitterText.parseTweet(tweet);
+            const canSubmit = !isSocialPosted && isTweetValid;
+
+            // We always render an error message, even if we don't display it.
```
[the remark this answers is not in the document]
EvanHahn: Removed in 49a640ace8194f0573b71dc43f3e631132661890.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
EvanHahn: Removed in 29b6726cd579ec1f7cf3a057498e91767102c869.

frontend/src/common/api/ClimateChangemakersAPI.ts
```diff
@@ -66,4 +67,11 @@ export const issueAPI = () =>
...
+}
```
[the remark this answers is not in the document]
EvanHahn: Done in 64919a7aded3327dffe4198d44ce28ec21fa9d73.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+                };
+                document.addEventListener('visibilitychange', onVisibilityChange, false);
+                const removeEventListener = document.removeEventListener.bind(document, 'visibilitychange', onVisibilityChange, false);
+
+                const timeout = setTimeout(done, 1000);
+                const clearWait = () => { clearTimeout(timeout); };
+            });
+
+            contents = (
+                <Form as="div">
```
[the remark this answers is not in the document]
EvanHahn: Done in 9705a527f2713f86e8cebdc4f6e96e7628ac7a3a.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,152 @@
...
+    preComposedTweet,
+    logTweet,
+}: Props) {
+    const [tweet, setTweet] = useState("");
+    const [hasClickedLink, setHasClickedLink] = useState(false);
+
+    useEffect(() => {
+        if (preComposedTweet.status === "loaded") {
+            setTweet(preComposedTweet.value);
+        }
```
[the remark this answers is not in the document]
EvanHahn: Done in 7fee34b3110da84860e008413c5d84094dc4f1da.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
EvanHahn: Done in b1009c3e50f00794a03ccfd2d7b84df840d9243e.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
EvanHahn: Removed it in 9705a527f2713f86e8cebdc4f6e96e7628ac7a3a.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
EvanHahn: Sidestepped in 9705a527f2713f86e8cebdc4f6e96e7628ac7a3a.

frontend/src/common/api/ClimateChangemakersAPI.ts
EvanHahn: Sidestepped this in 64919a7aded3327dffe4198d44ce28ec21fa9d73.

frontend/src/Pages/TakeAction/PostOnSocial.tsx
EvanHahn: Done in d374a422ee80c23cff3d2f2bbfc501b74cff703c.

jacknugent (changes requested):

frontend/src/Pages/TakeAction/PostOnSocial.tsx
```diff
@@ -1,15 +1,122 @@
...
+            if (openedWindow) {
+              setHasOpenedTwitter(true);
+            }
+        } else {
+            setIsSocialPosted(true);
+            const bioguideIds = actionInfo.legislators.map((l) => l.bioguideId);
+            let error: unknown;
+            try {
+                ({ error } = await logTweetAPI(actionInfo.initiatorEmail, selectedIssue.id, bioguideIds));
+            } catch (err: unknown) {
```
jacknugent, line 51: Can you just do:
```javascript
const response = await logTweetAPI(actionInfo.initiatorEmail, selectedIssue.id, bioguideIds));
if(response.error)
     console.warn(error);
```

frontend/src/Pages/TakeAction/TakeActionPage.tsx
```diff
@@ -26,6 +28,12 @@ export default function TakeActionPage() {
     useEffect(() => { isPhoneCallMade && scrollToId("post_on_social") }, [isPhoneCallMade])
     useEffect(() => { isSocialPosted && scrollToId("all_done") }, [isEmailSent, isPhoneCallMade, isSocialPosted])
 
+    const selectedIssueId = selectedIssue?.id;
+    const {
+        data: preComposedTweetData,
+        error: preComposedTweetError,
+    } = useSWR<{ tweet: string }, string>(() => (selectedIssueId !== undefined ? `/issues/${selectedIssueId}/precomposed-tweet` : null), fetcher);
```
jacknugent, line 35: Slightly simpler ternary syntax:
```javascript
!selectedIssueId ? null : `/issues/${selectedIssueId}/precomposed-tweet`
```

jacknugent (approved):
Just two small suggestions, otherwise looks good

frontend/src/Pages/TakeAction/TakeActionPage.tsx
jacknugent: Ok

frontend/src/Pages/TakeAction/PostOnSocial.tsx
EvanHahn: Done in 2bc151a4f4a414e2ede1d5fbae121aa16159979c.

EvanHahn:
Addressed all comments and merged `trunk`. I think this should be ready to go, though in the future we'll need to send a few more things to the precomposed tweet endpoint.

PR #113, opened 2021-10-10
Jnugent/send email

jacknugent:
- Add [use-swr](https://www.npmjs.com/package/swr) for improve data fetching
image


jacknugent:
> I'm not sure what the designs specify, but is it possible the avoid the whole `Command + Click` to select multiple? I think that's a pretty poor UX. A lot of people who will be using this won't know how to do that. As far as multi select dropdowns go, I think Github's "request reviewers" is a pretty decent example of dropdown multi selection. 

The Command + Click method is what the default HTML does. I'll have to use a library to do something like Github's. I can find something

kevincianfarini:
That looks great! I will pull and build momentarily

PR #127, opened 2021-10-14
Add bioguideIds to precomposed-tweet endpoint

jacknugent:
Also reformatted some files with prettier


kevincianfarini:
Closes #100

EvanHahn (approved):
LGTM but I'd recommend using `URLSearchParams` if we can.

PR #132, opened 2021-10-16
Redesign scrollspy

jacknugent:
Desktop:
image
Design:
image

Mobile:
image
Design:
image

- Introduction gets checked if you scroll past the section,
- You can click active/complete steps to scroll to them on desktop
- Removed "Write your why" step since we removed that from the app


frontend/src/Pages/TakeAction/ScrollSpy/Step.tsx
```diff
@@ -0,0 +1,86 @@
...
+                        "border-purple": !desktop && state !== "disabled",
+                        "border-disabled": !desktop && state === "disabled",
+                        "text-white": !desktop && state === "complete",
+                        "me-4 step-icon-desktop": desktop,
+                        "text-pink": desktop && state === "complete",
+                        "bg-dark-purple font-weight-800": state === "complete",
+                    })}
+                >
+                    {state === "complete" ? (
+                        <span>&#10003;</span>
```
EvanHahn, line 46: Could we do this with CSS instead, so users with screen readers don't even see the checkmark? I imagine applying a class and then including a `::before` element like this:

```css
/* not sure if this code is right */
.the-class::before {
  content: '✓';
  display: inline-block;
}
```

I left a similar comment above.

frontend/src/Pages/TakeAction/ScrollSpy/ScrollSpy.tsx
```diff
@@ -1,31 +1,79 @@
...
+        if (introSectionDistanceFromTop <= 0) setScrolledPastIntro(true);
+
+        handleResize();
+        window.addEventListener("scroll", handleResize);
+        window.addEventListener("resize", handleResize);
+        return () => {
+            window.removeEventListener("scroll", handleResize);
+            window.removeEventListener("resize", handleResize);
+        };
+    });
```
EvanHahn, line 31: I think we want to put some dependencies in here because I don't think we wanna re-run this effect on every render. I think we should also move `handleResize` into the effect.

frontend/src/Pages/TakeAction/ScrollSpy/ScrollSpy.tsx
```diff
@@ -1,31 +1,79 @@
...
+            })}
+        >
+            <Card.Body
+                className={cx("d-flex justify-content-between text-dark pt-3 pb-2 ps-0 pe-0", {
+                    "flex-column bg-dark-purple text-white": desktop,
+                })}
+            >
+                {desktop && (
+                    <h3 className="mb-0">
+                        <span className="step-icon-desktop ms-4 invisible">&#10003;</span>Steps
```
EvanHahn, line 50: Why is this invisible checkmark here? Do we wanna remove it? If not, could we use a `::before` in CSS to avoid putting it directly into the HTML? (See a similar comment below.)

frontend/src/Pages/TakeAction/ScrollSpy/Step.module.css
```diff
@@ -0,0 +1,25 @@
+.scrollSpyLine {
+    height: 2px !important;
```
EvanHahn, line 2: Does this only work with `!important`?

frontend/src/Pages/TakeAction/ScrollSpy/ScrollSpy.tsx
```diff
@@ -1,31 +1,79 @@
...
 
 type Props = {
     isEmailSent: boolean;
     isPhoneCallMade: boolean;
     isSocialPosted: boolean;
+    desktop?: boolean;
 };
 
-export default function ScrollSpy({ isEmailSent, isPhoneCallMade, isSocialPosted }: Props) {
+export default function MobileScrollSpy({ isEmailSent, isPhoneCallMade, isSocialPosted, desktop = false }: Props) {
```
EvanHahn, line 14: I left a high-level comment that might make this obsolete, but this is being used for both desktop and mobile. Maybe just rename this to `ScrollSpy`? (That's how it's used outside of this file, too.)

```suggestion
export default function ScrollSpy({ isEmailSent, isPhoneCallMade, isSocialPosted, desktop = false }: Props) {
```

frontend/src/Pages/TakeAction/ScrollSpy/ScrollSpy.tsx
```diff
@@ -1,31 +1,79 @@
...
+    const [introSectionDistanceFromTop, setIntroSectionDistanceFromTop] = useState(1);
+    const [scrolledPastIntro, setScrolledPastIntro] = useState(false);
+
+    const handleResize = () =>
+        setIntroSectionDistanceFromTop(document.getElementById("send_an_email")!.getBoundingClientRect()!.top);
+
+    useEffect(() => {
+        if (introSectionDistanceFromTop <= 0) setScrolledPastIntro(true);
```
EvanHahn, line 22: I wonder if this could be a bigger number. I'm on a monitor where `window.innerHeight === 1100` and I didn't see the Introduction step checked until I completed "Send an Email". I could see it just waiting for you to scroll at all, but maybe worth checking with a more design-minded person than me.

EvanHahn (approved):
Some high-level comments:

- I think what you have is fine, but I actually liked having a different desktop and mobile component for the scrollspy. There are so many conditionals in both `<ScrollSpy>` and `<Step>` that it might make sense to make them separate components again, even if it means creating helper components or having some repetition. That's just my opinion, though; what you have is fine.
- Nitpick: you say "desktop" and "mobile", but I think more accurate names would be things like "wide" or "narrow"...what do you think?

frontend/src/Pages/TakeAction/ScrollSpy/Step.module.css
jacknugent: Yes. If I remove it, it gets overridden by bootstrap's 1px height

frontend/src/Pages/TakeAction/ScrollSpy/ScrollSpy.tsx
jacknugent: This is cheating to add space to line up Steps with what's below. I removed it and added more padding instead.

frontend/src/Pages/TakeAction/ScrollSpy/Step.tsx
jacknugent: > Could we do this with CSS instead, so users with screen readers don't even see the checkmark? I imagine applying a class and then including a `::before` element like this:
> 
> ```css
> /* not sure if this code is right */
> .the-class::before {
>   content: '✓';
>   display: inline-block;
> }
> ```
> 
> I left a similar comment above.

Screenreaders won't pick up the checkmark at the wrong times because it is not in the dom until `state === "complete"`. I think ::before is less clear here because we are showing a number OR a checkmark, not a number that also has a ::before checkmark

jacknugent:
> Some high-level comments:
> 
>     * I think what you have is fine, but I actually liked having a different desktop and mobile component for the scrollspy. There are so many conditionals in both `<ScrollSpy>` and `<Step>` that it might make sense to make them separate components again, even if it means creating helper components or having some repetition. That's just my opinion, though; what you have is fine.

How about a compromise. I keep the components shared to allow code reuse (especially the state/logic, which is identical on mobile/desktop). To separate styling, I created a ternary condition for `desktop` that wraps all the JSX in `<Step>`, so there is _some_ duplicate code, but we get the benefit of a completely separate JSX for desktop/mobile in `<Step>` where the css was the craziest (I don't think it's _too_ bad in `<ScrollSpy>`). Let me know what you think.

>     * Nitpick: you say "desktop" and "mobile", but I think more accurate names would be things like "wide" or "narrow"...what do you think?

Using wide/narrow may be confusing because technically the Scrollyspy is more narrow on the wider screen. I could use WideScreen/NarrowScreen to clear that up, but that's a little more wordy. Thoughts?

frontend/src/Pages/TakeAction/ScrollSpy/ScrollSpy.tsx
EvanHahn: Here's what it looks like for me in Firefox:

https://user-images.githubusercontent.com/777712/137597704-cf7d18cb-2a27-400b-996a-e07126edea6f.mp4

Maybe we could `setScrolledPastIntro(true)` if we've scrolled even a little bit?

frontend/src/Pages/TakeAction/ScrollSpy/ScrollSpy.tsx
```diff
@@ -1,31 +1,83 @@
...
     isPhoneCallMade: boolean;
     isSocialPosted: boolean;
+    desktop?: boolean;
 };
 
-export default function ScrollSpy({ isEmailSent, isPhoneCallMade, isSocialPosted }: Props) {
+export default function ScrollSpy({ isEmailSent, isPhoneCallMade, isSocialPosted, desktop = false }: Props) {
+    const [introSectionDistanceFromTop, setIntroSectionDistanceFromTop] = useState(Math.min);
+    const [scrolledPastIntro, setScrolledPastIntro] = useState(false);
+    const introScrollBuffer = 40;
```
EvanHahn, line 17: I could see moving this to a top-level const (`INTRO_SCROLL_BUFFER` or something), but what you have here is fine.

EvanHahn (approved):

frontend/src/Pages/TakeAction/ScrollSpy/ScrollSpy.tsx
jacknugent: Oh I see what you mean. I mark as checked once the user passes the `<hr>` above "Send an email". That's how other scrollspy's do it, like in [Bootstrap](https://getbootstrap.com/docs/4.0/components/scrollspy/), a user hasn't made it to the next section until the previous one is out of view. Making it checked only after a little scrolling would be especially confusing on mobile, since the user mas 3 reps to scroll past.

To compromise, I added a `buffer` constant to the conditional, so that Intro is marked as checked 40 pixels above the `<hr>`, which is about where the user has scrolled past the reps. Let me know what you think

frontend/src/Pages/TakeAction/ScrollSpy/ScrollSpy.tsx
jacknugent: Moved up

PR #153, opened 2021-10-26
Implement new email design

### LINKED ISSUE #116: More inputs when sending email
Currently send email has smaller inputs (prefix, first name, last name, letter topic, subject) and one long input (body). I know Eliza mentioned maybe some other smaller inputs to help guide the user and personalize their email such as "Tell your rep where you're from" and "What do you do?"

Right now the design has room for them (notice the gap above "Draft Your Email")

image

Need to discuss potential options. Should we send these inputs to the back-end, or the front-end could simply append them to the body itself. Thoughts? @kevincianfarini

jacknugent:
Closes #116

image

image

image


EvanHahn:
Probably won't be able to review this for a few more days, just FYI.

jacknugent:
> In general I think this looks good. Some general comments on functionality and design after playing with it --
> 
>     * When getting to the email section, I see that there's three buttons in the pre-email drafting section. In the designs there's two buttons. This feels a bit awkward.
> 
>     * If I click "draft from scratch" the buttons become disabled and I can't go back to select the prompts section instead. In the event of a misclick, this could be really annoying.
> 
>     * In the "Your Email" section, the heading is kind of small. I see it's also designed this way in Figma. Part of me feels like this was an oversight. How do we feel about making that text styling the same as the other section headings? I personally feel that it's hard to discern between the prompts section and the review/draft from scratch section.
> 
>     * The draft with prompts section lacks the "Email Prompts" heading that shows up in Figma.
> 
>     * In the drafting from scratch text entry, the hint is still "write your why", which doesn't feel right. I see that's it's still that way in Figma as well, and I wonder if that's a side effect of copying and pasting design components.
> 
>     * Do we think it would be weird to have a required validator on the input fields of the prompts section? Maybe we shouldn't enable the "Review Email" button until at least one (or more) fields have content in them.
> 
> 
> Some of the above are discussion topics are some are design nitpicks. Feel free to reach out to me offline if you'd like to chat.

Per our chat today, I added the following:

“Draft from scratch” just moves you to Review email WITHOUT disabling anything, whereas clicking “Review email” moves you to Review email and disables the prompts (in other words you always have one chance to use the prompts to populate your review email)

“Your Email” same size as “Draft Your Email”

Add missing Email Prompts header from figma

Your Email placeholder becomes “Draft your email” instead of "Write your why..."

Instead of salutations input, add “To whom it may concern,” to the top of review your email after clicking “Review email”

Create input behind “Review email” that is required and gives message about “must fill out at least one prompt” — use HTML default validator

kevincianfarini (approved):

PR #154, opened 2021-10-29
Use PostgreSQL identity columns instead of serial

### LINKED ISSUE #74: Use identitiy columns instead of serial for Primary Keys

kevincianfarini:
Closes #74


database/ccm_schema.sql
```diff
@@ -616,19 +616,14 @@ CREATE TABLE public.action_contact_legislator (
 -- Name: action_contact_legislator_id_seq; Type: SEQUENCE; Schema: public; Owner: -
 --
 
-CREATE SEQUENCE public.action_contact_legislator_id_seq
+ALTER TABLE public.action_contact_legislator ALTER COLUMN id ADD GENERATED ALWAYS AS IDENTITY (
```
EvanHahn, line 619: This file looks autogenerated so I didn't review it. Lemme know if that's wrong.

backend/src/test/kotlin/org/climatechangemakers/act/feature/action/manager/DatabaseActionTrackerManagerTest.kt
```diff
@@ -14,7 +14,7 @@ class DatabaseActionTrackerManagerTest : TestContainerProvider() {
   private val manager = DatabaseActionTrackerManager(database, EmptyCoroutineContext)
 
   @Test fun `email action entries are recorded in both tables`() = suspendTest {
-    driver.insertIssue(1, "this is an issue", "tweet", "url.com")
+    val id = driver.insertIssue("this is an issue", "tweet", "url.com")
     driver.insertMemberOfCongress(DEFAULT_MEMBER_OF_CONGRESS.copy(bioguideId = "hello"))
     manager.trackActionSendEmail("foo@foo.com", "hello", 1)
```
EvanHahn, line 19: Do we need to update the last argument here?

```suggestion
    manager.trackActionSendEmail("foo@foo.com", "hello", id)
```

Same comment applies to some of the calls below.

backend/src/test/kotlin/org/climatechangemakers/act/feature/action/manager/DatabaseActionTrackerManagerTest.kt
```diff
@@ -36,12 +36,12 @@ class DatabaseActionTrackerManagerTest : TestContainerProvider() {
...
 
   @Test fun `recording a legislator call insert into both tables`() = suspendTest {
-    driver.insertIssue(1, "issue", "tweet", "url.com")
+    val id = driver.insertIssue("issue", "tweet", "url.com")
     driver.insertMemberOfCongress(DEFAULT_MEMBER_OF_CONGRESS.copy(bioguideId = "bioguide"))
     manager.trackActionPhoneCall("foo@foo.com", "bioguide", 1, "8675309")
 
     assertEquals(
-      1,
+      id,
```
EvanHahn, line 44: It seems weird that we'd check that the `COUNT(*)` is the same as the ID...should this be reverted?

EvanHahn (approved):
LGTM. Found a few things I didn't understand but I think things are fine.

database/ccm_schema.sql
kevincianfarini: This is indeed autogenerated!

backend/src/test/kotlin/org/climatechangemakers/act/feature/action/manager/DatabaseActionTrackerManagerTest.kt
kevincianfarini: yes, good call.

backend/src/test/kotlin/org/climatechangemakers/act/feature/action/manager/DatabaseActionTrackerManagerTest.kt
kevincianfarini: good catch!

PR #159, opened 2021-11-07
Include production docker build

### LINKED ISSUE #21: CI build that outputs a production container with static assets

kevincianfarini:
This commit adds a production Dockerfile to build a single backend
container that serves pre-built front end assets. This commit also adds
a CI task to build that container.

Closes #21

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,53 @@
+# build the client
+FROM node:lts-alpine3.14 AS build_client
+
+RUN mkdir -p /app
+COPY frontend/package*.json ./
+
+WORKDIR /app
+
+RUN npm install
+COPY frontend/ .
+
+RUN npm run build
+
+RUN mv build/ client/
+
+# build the backend
+
+FROM openjdk:11.0.12 AS build_backend
+
+RUN mkdir /appbuild
+COPY backend/ /appbuild
+
+RUN rm -rf /appbuild/src/main/resources/client
+RUN mkdir -p /appbuild/src/main/resources/client
+
+COPY --from=build_client /app/client/ /appbuild/src/main/resources/client
+
+WORKDIR /appbuild
+
+# TODO(kcianfarini) this is a bug in SQLDelight, the verifyMainDatabaseMigration task should not run.
+# sqldelight github issue 2654
+RUN ./gradlew clean build -x test -x verifyMainDatabaseMigration --no-daemon
+
+# Run the server
+
+FROM openjdk:11.0.12-jre
+
+ENV APPLICATION_USER 1033
+RUN useradd $APPLICATION_USER
+
+RUN mkdir /app
+RUN mkdir /app/resources
+RUN mkdir /app/log
+RUN chown -R $APPLICATION_USER /app
+RUN chmod -R 777 /app
+
+USER $APPLICATION_USER
+
+COPY --from=build_backend /appbuild/build/libs/climatechangemakers-backend*all.jar /app/climatechangemakers-backend.jar
+COPY --from=build_backend /appbuild/build/resources/ /app/resources/
+WORKDIR /app
+
+CMD ["sh", "-c", "java -server -XX:+UnlockExperimentalVMOptions -XX:+UseContainerSupport -XX:InitialRAMFraction=2 -XX:MinRAMFraction=2 -XX:MaxRAMFraction=2 -XX:+UseG1GC -XX:MaxGCPauseMillis=100 -XX:+UseStringDeduplication -jar climatechangemakers-backend.jar"]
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -0,0 +1,53 @@
+# build the client
+FROM node:lts-alpine3.14 AS build_client
```
kevincianfarini, line 2: @jacknugent you might consider updating the frontend Dockerfile to use this node docker image. Node recently updates and removed a bunch of hashing functions that Webpack relies are to generate stable file hashes when creating a build.

jacknugent: The front-end uses `alpine3.14`, how is that not the same as `lts-alpine3.14`?

backend/src/main/kotlin/org/climatechangemakers/act/feature/values/routing/ValuesRoutes.kt
```diff
@@ -3,9 +3,10 @@ package org.climatechangemakers.act.feature.values.routing
 import io.ktor.application.call
 import io.ktor.routing.Route
 import io.ktor.routing.get
+import io.ktor.routing.route
 import org.climatechangemakers.act.feature.values.controller.ValuesController
 
-fun Route.valuesRoutes(controller: ValuesController) {
+fun Route.valuesRoutes(controller: ValuesController) = route("/values") {
```
jacknugent, line 9: Why did this get added?

Dockerfile
kevincianfarini: alpine 3.14 is the linux version. lts refers to the node version I believe.

backend/src/main/kotlin/org/climatechangemakers/act/feature/values/routing/ValuesRoutes.kt
kevincianfarini: I removed it elsewhere to regain consistency in other places of the backend.

Dockerfile
jacknugent: If they both use alpine3.14, then what change are you suggesting to the frontend Dockerfile?

kevincianfarini: I _believe_ `node:alpine-3.14` tracks the latest version of node while running on alpine 3.14. In contrast, `node:lts-alpine3.14` tracks the lts version of node while running on alpine 3.14.

jacknugent: Ok, created #160

mcchang: can this dockerfile just depend on the image built from `frontend/Dockerfile` so that we don't have to duplicate the docker steps?

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN mv build/ client/
+
+# build the backend
+
+FROM openjdk:11.0.12 AS build_backend
+
+RUN mkdir /appbuild
+COPY backend/ /appbuild
+
+RUN rm -rf /appbuild/src/main/resources/client
```
mcchang, line 23: curious why this is necessary

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+# Run the server
+
+FROM openjdk:11.0.12-jre
+
+ENV APPLICATION_USER 1033
+RUN useradd $APPLICATION_USER
+
+RUN mkdir /app
+RUN mkdir /app/resources
```
mcchang, line 42: nit: could just run `mkdir -p /app/resources` and `mkdir -p /app/log` to create `/app` directory as necessary so that line 41 isn't necessary.

kevincianfarini: very true, didn't think about that.

Dockerfile
kevincianfarini: When I was working on this locally, I had a populated `client` directory which was causing conflicts. To ensure this directory is empty, I delete it during the build. 

Now that I think of it, this will almost always be run from CI so I guess it isn't strictly necessary.

kevincianfarini:
@mcchang 

> can this dockerfile just depend on the image built from frontend/Dockerfile so that we don't have to duplicate the docker steps?

It can't unfortunately, and this is because the `frontend/Dockerfile` also serves the assets via `npm start`.

mcchang:
> @mcchang
> 
> > can this dockerfile just depend on the image built from frontend/Dockerfile so that we don't have to duplicate the docker steps?
> 
> It can't unfortunately, and this is because the `frontend/Dockerfile` also serves the assets via `npm start`.
> 
> Edit: Maybe it can? I will look into it.

Hm, can't we just change this line from
```
COPY --from=build_client /app/client/ /appbuild/src/main/resources/client
```
to
```
COPY --from=build_client /app/build/ /appbuild/src/main/resources/client
```
or something like that? the `CMD` from the frontend dockerfile doesn't really matter for the purposes of doing a multi stage build

kevincianfarini:
@mcchang I will give it a go! I can follow up these comments in a separate PR. Probably tomorrow.

PR #157, opened 2021-11-03
Mock `check-membership` endpoint

kevincianfarini:
References #125

PR #165, opened 2021-11-16
Add signup step

### LINKED ISSUE #124: Include "join us" flow at the end

jacknugent:
Closes  #124


jacknugent:
image

frontend/src/Pages/TakeAction/ScrollSpy/ScrollSpy.tsx
```diff
@@ -75,11 +82,19 @@ export default function ScrollSpy({ isEmailSent, isPhoneCallMade, isSocialPosted
                     step={4}
                     id="#post_on_social"
                     state={linkState(isSocialPosted, isPhoneCallMade)}
-                    last
                     desktop={desktop}
                 >
                     {desktop ? "Post on Social" : "Post"}
                 </Step>
+                <Step
```
kevincianfarini, line 89: I think we should only conditionally show this step. If a call to check if the current user is already a changemaker returns true, we should omit this step.

jacknugent: Is there currently a call to check if the user is a changemaker?

kevincianfarini: Yes! A call to `/check-membership` will do it. Right now it's just mocked out to always return false. 

See more here https://github.com/climatechangemakers/climate-changemakers-act/pull/157

jacknugent: Fabulous, added

PR #173, opened 2021-11-19
Add Sincerely, {Name} to email body ending

### LINKED ISSUE #168: Include "Sincerely, {NAME}" at the end of an email drafted from prompts

jacknugent:
Closes #168

### DIFF
--- a/frontend/src/Pages/TakeAction/SendAnEmail/Prompts.tsx
+++ b/frontend/src/Pages/TakeAction/SendAnEmail/Prompts.tsx
@@ -1,6 +1,6 @@
 import HiddenValidationInput from "common/Components/HiddenValidationInput";
 import { EmailState } from "common/models/EmailState";
-import { useEffect, useState } from "react";
+import { useState } from "react";
 import { Button, Col, Form, Row } from "react-bootstrap";
 
 type Props = {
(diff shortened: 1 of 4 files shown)


frontend/src/Pages/TakeAction/SendAnEmail/Prompts.tsx
```diff
@@ -21,8 +23,10 @@ export default function Prompts({ formRef, setEmailState, setEmailBody }: Props)
     const [addedPrompts, setAddedPrompts] = useState(false);
 
     const addPromptsToEmail = () => {
-        if (formRef.current!.reportValidity()) {
-            setEmailBody([salutation, ...Object.values(emailPrompts).filter((o) => !!o)].join("\n\n"));
+        if (emailState === "reviewing" || formRef.current!.reportValidity()) {
+            setEmailBody(
+                [salutation, ...Object.values(emailPrompts).filter((o) => !!o), `Sincerely, ${firstName}`].join("\n\n")
```
kevincianfarini, line 28: can we add a newline here? 

> Sincerely,
> Kevin

jacknugent: We use [Prettier](https://prettier.io), which does the formatting for us, so I can't override this. We could tweak the average line length settings, but that would change other files. 

The idea of prettier is that it's very opinionated, but that's by design. The benefit is that it effectively ends all the personal opinions that go into JavaScript code bases. Read more [here](https://prettier.io/docs/en/option-philosophy.html)

kevincianfarini (approved):

PR #177, opened 2021-11-24
Simplify log tweeting

### LINKED ISSUE #174: Recording a Tweet is unintuitive
Last night at the first trial of the tool, we didn't collect any tweet events, which was strange. 

After some investigation and another user trial, I've determined it's because clicking "Send Tweet" doesn't log the event, but "Done Tweeting" does. Not a single person on our trial last night clicked "Done Tweeting", so no events were received. 

The fix to this is to record the action on "Send Tweet", and treat that button as the advancement to the next step (which means it will happen in the background while the user is interacting with Twitter).

jacknugent:
Before: You must click "Send Tweet" and then "Done Tweeting" to move to next step
Now: You just need to click "Send Tweet" to progress to next step

Closes #174

### DIFF
--- a/frontend/src/Pages/TakeAction/PostOnSocial.tsx
+++ b/frontend/src/Pages/TakeAction/PostOnSocial.tsx
@@ -27,24 +27,24 @@ export default function PostOnSocial({
     selectedIssue,
 }: Props) {
     const [tweet, setTweet] = useState(preComposedTweet);
+    const [hasOpenedTwitter, setHasOpenedTwitter] = useState(false);
 
     const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
         e.preventDefault();
         e.stopPropagation();
 
-        if (window.open(getPostTweetUrl(tweet!.trim()))) {
-            const response = await logTweetAPI(
-                actionInfo.initiatorEmail,
-                selectedIssue.id,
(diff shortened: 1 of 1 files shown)


jacknugent:
One issue -- with this change, we lose the smooth scroll to the next step, because after clicking "Send Tweet", we open a new tab, and then scrolling is disabled on the page. So when a user returns to the app, they are still on the Tweeting step, like this:
image

So the user will have to scroll down (to "Join Our Mission" currently) themselves. If this is a big concern, I could maybe try something hack-y like setting a .5 second timer when they return to the page which then scrolls them down, or maybe this method is preferred, since it might be jarring to have the page jump as soon as you return to it.

kevincianfarini:
@jacknugent maybe a good middle ground is to just have the log event when a user clicks `Send Tweet`, but to advance to the next step users must still explicitly press `Done Tweeting`.

jacknugent:
> @jacknugent maybe a good middle ground is to just have the log event when a user clicks `Send Tweet`, but to advance to the next step users must still explicitly press `Done Tweeting`.

This is the current behavior

jacknugent:
Done

EvanHahn (approved):

kevincianfarini (approved):

PR #181, opened 2021-12-01
Write sign up requests to Airtable CRM

### LINKED ISSUE #155: Include sign up endpoint

kevincianfarini:
Closes #155

### DIFF
--- a/backend/src/main/kotlin/org/climatechangemakers/act/feature/membership/model/AirtableCreateRecordRequest.kt
+++ b/backend/src/main/kotlin/org/climatechangemakers/act/feature/membership/model/AirtableCreateRecordRequest.kt
@@ -1,0 +1,53 @@
+package org.climatechangemakers.act.feature.membership.model
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import org.climatechangemakers.act.common.model.RepresentedArea
+import org.climatechangemakers.act.common.serializers.YesNoBooleanSerializer
+
+@Serializable class AirtableCreateRecordRequest private constructor(
+  val records: List<AirtableCreateRecord>,
+) {
+  constructor(
+    email: String,
+    firstName: String,
+    lastName: String,
+    city: String,
+    state: RepresentedArea,
+    experience: Boolean,
+    referral: String,
+  ) : this(
+    listOf(
+      AirtableCreateRecord(
+        AirtableCreateRecordFields(
+          email,
+          firstName,
+          lastName,
+          city,
+          state,
+          referral,
+          experience,
+        )
+      )
+    )
+  )
+}
+
+@Serializable class AirtableCreateRecord(
+  val fields: AirtableCreateRecordFields,
+)
+
+@Serializable class AirtableCreateRecordFields(
+  @SerialName("Email") val email: String,
+  @SerialName("First Name") val firstName: String,
+  @SerialName("Last Name") val lastName: String,
+  @SerialName("City") val city: String,
+  @SerialName("State") val state: RepresentedArea,
+  @SerialName("Referral?") val referral: String,
+
+  @SerialName("Experience?")
+  @Serializable(with = YesNoBooleanSerializer::class)
+  val experience: Boolean,
+) {
+  @SerialName("Country") val country = "USA"
+}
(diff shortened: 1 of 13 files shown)


backend/src/main/kotlin/org/climatechangemakers/act/feature/membership/model/AirtableCreateRecordRequest.kt
```diff
@@ -0,0 +1,53 @@
+package org.climatechangemakers.act.feature.membership.model
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import org.climatechangemakers.act.common.model.RepresentedArea
+import org.climatechangemakers.act.common.serializers.YesNoBooleanSerializer
+
+@Serializable class AirtableCreateRecordRequest private constructor(
+  val records: List<AirtableCreateRecord>,
```
EvanHahn, line 9: Airtable lets you create between 1 and 10 records in a single request (unless it's changed since I departed). Do we want to do anything special to account for that?

backend/src/main/kotlin/org/climatechangemakers/act/feature/membership/model/AirtableCreateRecordRequest.kt
```diff
@@ -0,0 +1,53 @@
...
+    )
+  )
+}
+
+@Serializable class AirtableCreateRecord(
+  val fields: AirtableCreateRecordFields,
+)
+
+@Serializable class AirtableCreateRecordFields(
+  @SerialName("Email") val email: String,
```
EvanHahn, line 41: The Airtable API has some brittleness here, because someone could rename the field and your API integration will break. One solution: use an undocumented but supported feature that lets you pass the field ID instead of its name. You'll have to poke around the devtools to get it (they start with `fld`).

It has the disadvantage of being brittle another way: someone can't replace that field because the ID will be different.

Up to you if you want to do this, and lemme know if I can help you find the field IDs.

EvanHahn (commented):
Partial review.

backend/src/main/kotlin/org/climatechangemakers/act/feature/membership/model/AirtableCreateRecordRequest.kt
kevincianfarini: I don't think so. That would require some kind of caching/queuing mechanism on the backend, which I think adds too much complexity for now.

backend/src/main/kotlin/org/climatechangemakers/act/feature/membership/model/AirtableCreateRecordRequest.kt
kevincianfarini: Yeah, I definitely thought of that. Overall, I'm pretty unpleased with Airtable. 

To your point, I will be communicating to the people who own this airtable service that if they rename tables it will break the action tool. We could look into doing it via undocumented fields, but that also feels fragile to me by nature of it being undocumented. 

There's even more things that worry me. 

Right now, the sign up flow will do two things: 

1. Sign the user up in the Airtable database 
2. Send a mailchimp "welcome" email which allows them to join slack, receive the newsletter, etc. 

The way our data is currently hosted makes it difficult to ensure data integrity. 

What happens if we successfully make the HTTP request to sign a user up, but mailchimp fails? How do we roll that sign up back? How do we indicate to the person signing up that it...kind of worked? 

There are ways, but none of them are elegant and they are all prone to rate limiting and other oddities with the Airtable API. 

Hosting all of this data in our own database makes this all much easier. It would: 

* Not be rate limited
* Allow transactions for data integrity. This is super important because it would allow us to do something like: 

```kt
database.transaction {
  database.signPersonUp(...)
  mailchimp.sendEmail(...)
}
```

In the above, if _either_ of the function calls within the transaction block fails, the transaction would be rolled back. 

tl;dr I think the solution to the myriad of problems we might experience with the Airtable API would be better solved by moving our data into a database we own versus introducing any queuing, batching, caching, or hacked rollback mechanisms. I plan to bring this up to the product folk at Climate Changemakers.

mcchang: Alternatively, we might want to host our customer data in a real CRM like salesforce or hubspot. It will be easier to integrate customer data with other external services if it's in a CRM as opposed to "just in a database"

EvanHahn (approved):
Approving without a full review and can do a follow up later.

PR #192, opened 2021-12-04
Replace join with amplify

### LINKED ISSUE #184: Replace city with zipcode on signup
Change necessary after #183

### LINKED ISSUE #188: Replace Join step with Amplify step, and move Join step to end

jacknugent:
- Create TwitterForm component to reuse logic for tweeting since we do two tweets now
- Add extra form elements for signup
- Move signup to end
- Add amplify section

Closes #184, #188, and #140


kevincianfarini:
Do we want "amplify your voice" to be a step in the progress tracker? Or is it supposed to be lumped in with "post on social"? 

If it's supposed to be lumped in, then we shouldn't mark "post on social" as completed until both of those actions are completed.

frontend/src/Pages/TakeAction/Amplify/Amplify.tsx
```diff
@@ -0,0 +1,46 @@
...
+    };
+
+    return (
+        <div className="pt-2 pb-3">
+            <div className="d-flex">
+                <img src={socialIcon} alt="" height="40" width="40" />
+                <h2 className="text-pink fw-bold mb-3 ms-3">Amplify Your Voice</h2>
+            </div>
+            <TwitterForm
+                preComposedTweet="Just used the @climatevote outreach tool to contact Congress, check it out!"
```
kevincianfarini, line 35: This should probably have a link to `https://act.climatechangemakers.org`, right?

jacknugent: Sure, added

jacknugent:
> Do we want "amplify your voice" to be a step in the progress tracker? Or is it supposed to be lumped in with "post on social"?
> 
> If it's supposed to be lumped in, then we shouldn't mark "post on social" as completed until both of those actions are completed.
> 
> EDIT: I think this is actually a bug. My suspicion is that you changed the label for "Join Our Mission" or whatever it was previously called to be "Amplify" without updating the logic that controls whether or not it's shown. "Amplify" is shown when using the tool from an email that's not already signed up to be a changemaker, but doesn't show up when I use when that's already part of our community.

It was a bug that Amplify was not showing up when a user is part of the community (I pushed a change to fix that). 

As for the "Post on Social" and "Amplify" steps, they are both in the figma:
image

and "Join Our Mission" is no longer a step (it's at the bottom of the page when you're finished). 

Do we want to make Post on Social and Amplify the same step? Or just rename Post on Social to something more specific, like "Tweet Your Reps"

kevincianfarini:
@jacknugent nope! It was just the bug that was confusing me.

kevincianfarini (approved):

PR #217, opened 2021-12-16
Include URL path while logging redacted requests

### LINKED ISSUE #199: Make log redaction more permissive
We should be including the whole path of an API endpoint we're hitting. We should only redact query params and api keys.

kevincianfarini:
For diagnostic purposes, it's useful to have the path of the request available in the logs. This commit still ensures that sensitive information included in the URL (such as username, password, and query params) are still redacted. 

Closes #199


kevincianfarini:
Admin merging for now so that I can make a deployment tonight. Will follow up reviews later.

PR #222, opened 2021-12-17
Tag generated docker images with git hash

kevincianfarini:


.gitlab-ci.yml
```diff
@@ -32,7 +32,7 @@ build-production-container:
   image: docker:stable
   stage: build
   script: 
-    - docker build -t $IMAGE_TAG .
+    - docker build -t $IMAGE_TAG:$(git rev-parse --short HEAD) .
```
EvanHahn, line 35: Could be worth quoting this whole thing to prevent word splitting. I'm reviewing on my phone but it's something like:

    "$IMAGE_TAG:$(...)"

It should be fine without it, but if either expression contains spaces, bad things could happen.

EvanHahn (approved):
LGTM other than one small nitpick.

mcchang:
Can we use the provided `CI_COMMIT_SHORT_SHA` env var instead of manually generating the SHA? https://docs.gitlab.com/ee/ci/variables/predefined_variables.html

.gitlab-ci.yml
```diff
@@ -41,7 +41,7 @@ publish-production-container:
   stage: build
   script: 
     - docker login -u $GITLAB_USER_LOGIN -p $GITLAB_API_KEY registry.gitlab.com 
-    - docker build -t $IMAGE_TAG .
+    - docker build -t $IMAGE_TAG:$(git rev-parse --short HEAD) .
     - docker push $IMAGE_TAG
```
mcchang, line 45: hm, we don't need to push the specific tag here or all tags? https://docs.docker.com/engine/reference/commandline/push/#push-all-tags-of-an-image

kevincianfarini: Yes im into that

kevincianfarini: https://blog.container-solutions.com/tagging-docker-images-the-right-way

.gitlab-ci.yml
kevincianfarini: Done!

kevincianfarini:
@mcchang 

>Can we use the provided CI_COMMIT_SHORT_SHA env var instead of manually generating the SHA? https://docs.gitlab.com/ee/ci/variables/predefined_variables.html

Done!

PR #225, opened 2021-12-19
Use useSWRImmutable to reduce Airtable calls

### LINKED ISSUE #224: Airtable showing a lot of request data
I took a look at the logs today, and it seems like something is awry causing endless requests to AirTable. Based on the time distribution of these requests, I highly doubt its origination is from the backend.

image

I will research this more tomorrow. I suspect this has something to do with the frontend.

jacknugent:
Closes #224

Read more about `useSWRImmutable` [here](https://swr.vercel.app/docs/revalidation)


kevincianfarini (approved):
Thanks for the fix!

PR #239, opened 2022-01-10
Update production container /etc/hosts for HCWC

### LINKED ISSUE #236: Set up request redirection through stable EC2 instance for Congressional APIs
We should first research how we might do this by leveraging `hosts` file redirects.

kevincianfarini:
House Communicating with Congress API requires a whitelisted
IP address. Therefore, Climate Changemakers requires that we have
infrastructure stable enough to always send reqeusts to HCWC from the
same IP address.

This commit ammends the `/etc/hosts` file of our production container so
that requests going to `cwc.house.gov` get proxied through a stable EC2
instance.

Closes #236

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,7 @@ WORKDIR /app
 ARG nginx_proxy_host
 
 RUN echo "echo ${nginx_proxy_host} cwc.house.gov >> /etc/hosts" >> /tmp/update_hosts.sh
+RUN echo "echo ${nginx_proxy_host} soapbox.senate.gov >> /etc/hosts" >> /tmp/update_hosts.sh
 RUN chmod 777 /tmp/update_hosts.sh
 
 USER root
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,4 +47,10 @@ COPY --from=build_backend /appbuild/build/libs/climatechangemakers-backend*all.j
 COPY --from=build_backend /appbuild/build/resources/ /app/resources/
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


kevincianfarini:
@mcchang we can talk offline about the intricacies of this PR. I want to confirm with you that I have the correct entry set up in our hosts file.

Dockerfile
```diff
@@ -47,4 +47,10 @@ COPY --from=build_backend /appbuild/build/libs/climatechangemakers-backend*all.j
 COPY --from=build_backend /appbuild/build/resources/ /app/resources/
 WORKDIR /app
 
-CMD ["sh", "-c", "java -server -XX:+UnlockExperimentalVMOptions -XX:+UseContainerSupport -XX:InitialRAMFraction=2 -XX:MinRAMFraction=2 -XX:MaxRAMFraction=2 -XX:+UseG1GC -XX:MaxGCPauseMillis=100 -XX:+UseStringDeduplication -jar climatechangemakers-backend.jar"]
+ARG nginx_proxy_host
```
mcchang, line 50: nit: would name this something like `cwc_proxy_host` to be more descriptive and specific as to what this arg points to

mcchang (commented):
seems reasonable! So this works when you run your container locally?

Dockerfile
kevincianfarini: yes, absolutely. I will change the environment variable in the pipeline, too.

kevincianfarini:
>  on fargate as well?

This is something I was hoping to confirm with you in a pairing session @mcchang. I will follow up with the CWC team to ensure they whitelisted the proper IP.

PR #264, opened 2022-01-21
Retain eips from public private vpc on networking stack deletion

mcchang:
We'd like to keep these EIPs around so that we can assign one of them to our SCWC and HCWC proxy box https://us-west-2.console.aws.amazon.com/ec2/v2/home?region=us-west-2#InstanceDetails:instanceId=i-0af64e54f5af0e1d5, since these two EIPs are whitelisted source ips from the HCWC perspective.


kevincianfarini (approved):
Looks fine to me!

PR #268, opened 2022-01-26
Deploy latest to AWS via CI

kevincianfarini:


jacknugent (approved):

kevincianfarini:
I originally wanted our pipeline to do a dry run of the cloudformation infrastructure, but it was too annoying and also exposed our aws keys in the pipeline. Womp womp.

.gitlab-ci.yml
```diff
@@ -67,3 +68,21 @@ migrate-production-database:
...
+  only: 
+    - trunk
+
+deploy-cloudformation-network:
+  stage: deploy
+  image: amazon/aws-cli:2.4.14
+  script: 
+    - aws cloudformation deploy --template-file cloudformation/public-vpc.yml --stack-name networking-public-vpc
+  only: 
+    - trunk
```
kevincianfarini, line 88: Let me know what you think of this @mcchang!

.gitlab-ci.yml
```diff
@@ -67,3 +68,24 @@ migrate-production-database:
...
+deploy-cloudformation-network:
+  stage: deploy
+  image: amazon/aws-cli:2.4.14
+  script: 
+    - aws cloudformation deploy --template-file cloudformation/public-vpc.yml --stack-name networking-public-vpc
+  only: 
+    refs:
+      - trunk
+    changes:
+      - cloudformation/public-vpc.yml
```
kevincianfarini, line 91: Let me know what you think @mcchang!

.gitlab-ci.yml
```diff
@@ -67,3 +68,24 @@ migrate-production-database:
       - trunk
     changes:
       - backend/src/main/sqldelight/migrations/*.sqm
+
+deploy-cloudformation-service:
+  stage: deploy
+  image: amazon/aws-cli:2.4.14
+  variables:
+    IMAGE_URL: "registry.gitlab.com/kevincianfarini/climate-changemakers-act:$CI_COMMIT_SHORT_SHA"
```
mcchang, line 76: this is fine, but curious is there's a way to get (at least part of) the image url from a gitlab env variable

.gitlab-ci.yml
mcchang: is it possible to specify in this yaml file that `deploy-cloudformation-service` depends on `deploy-cloudformation-network`? the `public-subnet-public-loadbalancer.yml` cloudformation stack depends on `public-vpc.yml` being up to date.

mcchang: not sure about gitlab / cloudformation best practices, but having this clause here might cause issues down the line where we don't deploy `public-vpc` even though it needs to be.  For example, `public-vpc` could depend on some other cloudformation stack that has changed, and needs to be redeployed even if `public-vpc.yml` itself has not been changed.

kevincianfarini: > is it possible to specify in this yaml file that deploy-cloudformation-service depends on deploy-cloudformation-network? the public-subnet-public-loadbalancer.yml cloudformation stack depends on public-vpc.yml being up to date.

Yes I believe so. I will make that improvement. 

> not sure about gitlab / cloudformation best practices, but having this clause here might cause issues down the line where we don't deploy public-vpc even though it needs to be. For example, public-vpc could depend on some other cloudformation stack that has changed, and needs to be redeployed even if public-vpc.yml itself has not been changed.

We could potentially alter this clause if we get another yaml file that this depends on? That way, everything would deploy!

.gitlab-ci.yml
kevincianfarini: Let me look into this!

.gitlab-ci.yml
```diff
@@ -67,3 +67,29 @@ migrate-production-database:
...
+deploy-cloudformation-service:
+  stage: deploy
+  image: 
+    name: amazon/aws-cli:2.4.14
+    entrypoint: [""]
+  variables:
+    IMAGE_URL: "registry.gitlab.com/kevincianfarini/climate-changemakers-act:$CI_COMMIT_SHORT_SHA"
+  script: 
+    - aws cloudformation deploy --template-file cloudformation/public-subnet-public-loadbalancer.yml --stack-name climate-changemakers-act --parameter-overrides ImageUrl=$IMAGE_URL
+  needs: ["deploy-cloudformation-network", "migrate-production-database"]
```
kevincianfarini, line 93: @mcchang this line ensures that `deploy-cloudformation-service` will happen after both `deploy-cloudformation-network` and `migrate-production-database` complete. 

Both of the build jobs listed in this needs clause are allowed to run in parallel, but `deploy-cloudformation-service` will not start until both are complete. 

> not sure about gitlab / cloudformation best practices, but having this clause here might cause issues down the line where we don't deploy public-vpc even though it needs to be. For example, public-vpc could depend on some other cloudformation stack that has changed, and needs to be redeployed even if public-vpc.yml itself has not been changed.

I think we should punt on this. It should be the responsibility of a PR creator to update the triggers under which `deploy-cloudformation-network` job is launched. This would likely consist of just adding another file to the `changes` clause. 

Thoughts?

.gitlab-ci.yml
kevincianfarini: Done!

.gitlab-ci.yml
kevincianfarini: I'm going to merge for now! Will introduce follow ups if necessary.

mcchang: punting sounds good, just seems strange that we opt to special case `deploy-cloudformation-network` on changes to `cloudformation/public-vpc.yml` when we don't do the analogous thing for `deploy-cloudformation-service` and `cloudformation/public-subnet-public-loadbalancer.yml`.  Just a timesaver?

kevincianfarini: Yes. On every single deploy we will have a new docker container published that's worthy of publishing. 

That's not the case with the networking stack.

PR #274, opened 2022-02-02
Add gender-neutral text

### LINKED ISSUE #269: Indicate to users that gender neutral options for the email form aren't available because of Congress
We should allow they/them but congress ain't woke!

jacknugent:
Closes #269 

image


kevincianfarini:
I wanted to chat with you before you got to this. Can this be a tooltip next to the field?

kevincianfarini:
Follow up: This should be a tooltip inline with `Your Prefix`. I've pinged the designers about an icon we can use for the tooltip mouseover.

kevincianfarini:
Here's an icon recommended by Elizabeth for the tooltip! https://fontawesome.com/v5.15/icons/info-circle?style=solid

jacknugent:
image

Added font-awesome, which I'd love to keep using for future icons :)

kevincianfarini:
This looks awesome now, thank you!

kevincianfarini (approved):

PR #275, opened 2022-02-02
Add default email if drafting from scratch

### LINKED ISSUE #273: Users who "skip to call" are presented with a blank call script.
> Hello, I’m [NAME], a constituent calling from [LOCATION].
> 
> As a [INTRODUCE YOURSELF: WHERE DO YOU LIVE? WHAT DO YOU DO? WHAT’S IMPORTANT TO YOU?], I am concerned about the climate crisis because [WHY YOU CARE ABOUT CLIMATE].
> I’m calling to urge [Senator/Representative NAME] to take action by [CLEAR ASK].
> This policy would [WHY IS THE POLICY IMPORTANT].
> Now is the time for [Senator/Representative NAME] to address these pressing issues by [REPEAT CLEAR ASK].
> Please pass along my concerns to the [Senator/Representative]. Thanks so much for taking my call, and please thank the [Senator/Representative] for their service and leadership.

jacknugent:
Closes #273 

image

Took the liberty of adding first name instead of [Name]


kevincianfarini:
Two things: 

1. This was originally supposed to go into the call script field if it was blank upon making calls. 
2. I _think_ this isn't an issue for right now. We were finalizing some edits to the cool and wrapping loose ends, but the meeting ran long and will be finished on Thursday. I will keep you posted.

kevincianfarini:
This is a go! However, this isn't for emails. This only applies if a user skips emailing altogether and wants to make calls. We want to provide them with _something_ so they're not making the calls blind.

jacknugent:
Ok moved the text to make a call section ONLY if no email body

kevincianfarini (approved):

PR #281, opened 2022-02-08
Enroll changemakers in our newsletter

### LINKED ISSUE #34: Start enrolling changemakers into our promotional emails
If a changemaker decides to opt-in to our promotional emails in the action tool, we need to register them in our database as such. Currently, the `contacts` table contains various outreach datapoints. The `mailchimp_status` seems like it could be relevant, but running the following query produces lackluster results. 

```sql
SELECT DISTINCT mailchimp_status FROM contacts;

+--------------------+
| mailchimp_status   |
|--------------------|
| <null>             |
+--------------------+
```

There's a need to understand how email addresses are currently collected for promotional outreach. One possible solution is to drop the `mailchimp_status` column and replace it with the following column. 

| Column Name | Type | Modifiers | Default |
| -- | -- | -- | -- |
| promotional_enrolled| boolean | NOT_NULL | false |

kevincianfarini:
Closes #34

### DIFF
--- a/backend/src/main/kotlin/org/climatechangemakers/act/feature/action/controller/ActionController.kt
+++ b/backend/src/main/kotlin/org/climatechangemakers/act/feature/action/controller/ActionController.kt
@@ -36,10 +36,8 @@ class ActionController @Inject constructor(
     val request = call.receive<InitiateActionRequest>()
 
     val response: Deferred<InitiateActionResponse> = coroutineScope {
-      launch {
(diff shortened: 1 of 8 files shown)


EvanHahn:
I can take a look on Thursday if nobody gets to it before then.

kevincianfarini:
This was a long open issue that I could now address since the infrastructure for interacting with Mailchimp was addressed in #196

jacknugent (approved):

PR #286, opened 2022-02-10
Record Congressional Email Delivery ID

### LINKED ISSUE #262: `action_email_legislator` should be storing email delivery IDs

kevincianfarini:
We should be storing these IDs for debugging purposes. In the event that
an email doesn't get delivered where it's supposed to (looking at you
Geocod.io), we need to be able to hunt down what happened.

Closes #262

### DIFF
--- a/backend/src/main/sqldelight/migrations/V3__add_email_delivery_id.sqm
+++ b/backend/src/main/sqldelight/migrations/V3__add_email_delivery_id.sqm
@@ -1,0 +1,2 @@
+ALTER TABLE action_email_legislator
+ADD COLUMN email_delivery_id VARCHAR;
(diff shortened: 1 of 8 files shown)


backend/src/main/sqldelight/migrations/V3__add_email_delivery_id.sqm
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE action_email_legislator
+ADD COLUMN email_delivery_id VARCHAR;
```
kevincianfarini, line 2: unfortunately this needs to be nullable for the time being. I didn't have the foresight to add this in from the beginning, and there's no way for us to backfill this data.

jacknugent (approved):

PR #289, opened 2022-02-13
Fix email retry strategy

### LINKED ISSUE #285: Failed bioguide IDs not being respected when email delivery fails
In development, when I try to send an email and one of the recipient fails, I would expect retries only be conducted for the failed bioguide. Instead, I'm seeing that the `failedBioguideIds` in the error response body aren't respected. For example, this is the result of a single send email attempt which fails. 

```
postgres@0:postgres> select * from action_contact_legislator;
+------+---------+-------------------------------+------------+-------------------------+
| id   | email   | contacted_at                  | issue_id   | contacted_bioguide_id   |
|------+---------+-------------------------------+------------+-------------------------|
| 1    | k@c.com | 2022-02-10 01:55:19.674162+00 | 7          | K000384                 |
| 3    | k@c.com | 2022-02-10 01:55:29.607386+00 | 7          | W000805                 |
| 4    | k@c.com | 2022-02-10 01:55:29.649851+00 | 7          | K000384                 |
| 5    | k@c.com | 2022-02-10 01:55:39.641468+00 | 7          | K000384                 |
| 6    | k@c.com | 2022-02-10 01:55:39.641737+00 | 7          | W000805                 |
| 7    | k@c.com | 2022-02-10 01:55:49.588589+00 | 7          | K000384                 |
| 8    | k@c.com | 2022-02-10 01:55:49.616595+00 | 7          | W000805                 |
| 2    | k@c.com | 2022-02-10 01:55:19.674471+00 | 7          | W000805                 |
+------+---------+-------------------------------+------------+-------------------------+
```

kevincianfarini:
Closes #285


frontend/src/Pages/TakeAction/SendAnEmail/SendAnEmail.tsx
```diff
@@ -10,7 +10,7 @@ import { Issue } from "common/models/Issue";
 import { useEffect, useRef, useState } from "react";
 import { Accordion, Button, Col, Form, OverlayTrigger, Row, Tooltip } from "react-bootstrap";
 import Select from "react-select";
-import useSWRImmutable from "swr";
+import useSWRImmutable from "swr/immutable";
```
kevincianfarini, line 13: @jacknugent I snuck this in too. This was that swr immutable namespace bug we ran into a while ago.

frontend/src/common/api/ClimateChangemakersAPI.ts
```diff
@@ -57,17 +57,6 @@ const post = async <Data, Error = string>(path: string, content: Object): Promis
...
-    let response = await fetch();
-    if (!response.successful) {
-        for (var i = 0; i < 3; i++) {
-            response = await fetch();
-            if (response.successful) return response;
-        }
-    }
-    return response;
-};
-
```
kevincianfarini, line 70: I saw retries happening when I tested this. How exactly are you achieving that without this function?

jacknugent: Retries were happening, but `contactedBioguideIds` never changes on retries, so it'll always retry the same three, even if some succeed. Now I'm checking and changing `contactedBioguideIds` on line 108

jacknugent: Retries were happening, but contactedBioguideIds never changes on retries, so it'll always retry the same three, even if some succeed. Now I'm checking and changing contactedBioguideIds on line 108

jacknugent (approved):

PR #312, opened 2022-03-24
Use Google Civic Information API

kevincianfarini:
This commit replaces Geocod.io as our source of congressional district
information. This is important because Geocod.io has a lot of
informational errors. Emails that are sent to an incorrect district are
rejected by CWC.

Unfortunately the Google Civic Information API does not expose GPS coordinates.
This is a separate, paid, service from Google.

Therefore, we now have two GIS dependencies. Google's will hopefully
give us correct congressional district data, and Geocodio will give us
GPS coordinates that we will continue to use for district office
lookups.

### DIFF
--- a/backend/src/main/kotlin/org/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
+++ b/backend/src/main/kotlin/org/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
@@ -1,7 +1,7 @@
 package org.climatechangemakers.act.feature.findlegislator.manager
 
 import org.climatechangemakers.act.feature.findlegislator.model.GetLegislatorsByAddressRequest
-import org.climatechangemakers.act.feature.findlegislator.model.Legislator
+import org.climatechangemakers.act.feature.findlegislator.model.MemberOfCongressDto
 import org.climatechangemakers.act.feature.findlegislator.model.LegislatorArea
 import org.climatechangemakers.act.feature.findlegislator.service.GeocodioService
 import org.climatechangemakers.act.feature.lcvscore.manager.LcvScoreManager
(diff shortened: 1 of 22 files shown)


kevincianfarini:
I can't merge this until I include the new Google civic api key in AWS

backend/src/main/kotlin/org/climatechangemakers/act/feature/findlegislator/service/GoogleCivicService.kt
```diff
@@ -0,0 +1,16 @@
...
+import retrofit2.http.Query
+
+interface GoogleCivicService {
+
+  @GET("representatives")
+  suspend fun getCongressionalDistrict(
+    @Query("address") address: String,
+    @Query("levels") levels: String = "country",
+    @Query("roles") roles: String = "legislatorLowerBody",
+    @Query("includeOffices") includeOffices: Boolean = false,
```
EvanHahn, line 14: I might've missed it in my quick review, but are these other parameters used?

backend/src/main/kotlin/org/climatechangemakers/act/feature/findlegislator/model/GoogleCivicApiResult.kt
```diff
@@ -0,0 +1,16 @@
+package org.climatechangemakers.act.feature.findlegislator.model
+
+import kotlinx.serialization.Serializable
+
+private val openCivicDataRegex = """ocd-division/country:us/state:[a-zA-Z]{2}/cd:(\d+)""".toRegex()
+
+@Serializable class GoogleCivicApiResult(
+  private val divisions: Map<String, Map<String, String>>,
+) {
+  val congressionalDistrict: Short get() = openCivicDataRegex
```
EvanHahn, line 10: Are we sure we want this to be a `Short`? Is it worth picking a larger data type?

EvanHahn (approved):
Did a cursory review on my phone. Looks good!

It's a whole can of worms, but I wonder if we could cache some of this stuff to cut down on costs. Probably not worth it because it's looking up by address, but something I'm thinking about...

kevincianfarini:
>It's a whole can of worms, but I wonder if we could cache some of this stuff to cut down on costs. Probably not worth it because it's looking up by address, but something I'm thinking about...


Our HTTP library will cache GET requests that are identical for some time. I don't think we should do any caching aside from that.

backend/src/test/kotlin/org/climatechangemakers/act/feature/findlegislator/model/GoogleCivicApiResultTest.kt
```diff
@@ -0,0 +1,25 @@
...
+
+class GoogleCivicApiResultTest {
+
+  @Test fun `civic api result can pase single digit congressional division from open civic data format`() {
+    val formatted = "ocd-division/country:us/state:va/cd:4"
+    val actual = GoogleCivicApiResult(mapOf(formatted to emptyMap())).congressionalDistrict
+    assertEquals(4, actual)
+  }
+
+  @Test fun `civic api result can pase single digit congressional division leading zero from open civic data format`() {
```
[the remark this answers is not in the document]
kevincianfarini: yes

backend/src/main/kotlin/org/climatechangemakers/act/feature/findlegislator/manager/LegislatorsManager.kt
```diff
@@ -10,29 +10,33 @@ import kotlinx.coroutines.async
...
-      async {
-        val bioguideId = geocodioLegislator.references.bioguide
-        val lcvScores = async { getLcvScoresForBioguide(bioguideId) }
-        val memberOfCongress = async { memberOfCongressManager.getMemberOfCongressForBioguide(bioguideId) }
-        val districtPhoneNumber = async {
-          districtOfficerManager.getNearestDistrictOfficePhoneNumber(bioguideId, geoCodioResponse.location)
-        }
+  suspend fun getLegislators(request: GetLegislatorsByAddressRequest): List<MemberOfCongressDto> = coroutineScope {
+    val membersOfCongress = async {
+      val googleCivicResponse = googleCivicService.getCongressionalDistrict(request.queryString)
```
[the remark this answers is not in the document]
kevincianfarini: request.querystring is a synthetic property that's built from the components of `GetLegislatorsByAddressRequest`. 

```kt
private val GetLegislatorsByAddressRequest.queryString: String get() = "$streetAddress, $city $state $postalCode"
```

backend/src/main/kotlin/org/climatechangemakers/act/feature/findlegislator/service/GoogleCivicService.kt
kevincianfarini: These aren't ever used as parameters, but these are required to pare down the response body to just the information we need. 

I can look into a better way to do it that isn't exposed in our request API.

backend/src/main/kotlin/org/climatechangemakers/act/feature/findlegislator/model/GoogleCivicApiResult.kt
kevincianfarini: Short has a range of -32,768 to 32,767. We could potentially even go smaller with a byte. I don't think any state has any congressional district number above 127. 

Eg. California has 53 congressional districts.

EvanHahn: Cool, that makes sense. Just wanted to double-check!

PR #320, opened 2022-05-01
Add null constraint to `district_office.phone_number`

kevincianfarini:
The only piece of information which is useful to Changemaker's is the
district office's phone number. To aid in the automatic processes for
updating this data, this commit makes that column non-null.

References #317

### DIFF
--- a/backend/src/main/sqldelight/migrations/V7__non_null_district_office_phone_number.sqm
+++ b/backend/src/main/sqldelight/migrations/V7__non_null_district_office_phone_number.sqm
@@ -1,0 +1,6 @@
+DELETE FROM district_office
+WHERE phone_number IS NULL;
+
+ALTER TABLE district_office
+ALTER COLUMN phone_number
+SET NOT NULL;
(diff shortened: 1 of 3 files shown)


backend/src/main/sqldelight/migrations/V7__non_null_district_office_phone_number.sqm
```diff
@@ -0,0 +1,6 @@
+DELETE FROM district_office
+WHERE phone_number IS NULL;
```
EvanHahn, line 2: We sure this is okay? (I think so, just wanna double check.)

EvanHahn (approved):

backend/src/main/sqldelight/migrations/V7__non_null_district_office_phone_number.sqm
kevincianfarini: Can you elaborate your concern? 

I tested this again my local development database and it worked. `district_office` doesn't have any table references, so the rows we're deleting won't cause any cascades.

EvanHahn: My main concern was "are there any district offices that we don't want to delete"? I think the answer is "nope, everything is fine", but I wanted to be careful when possibly deleting important data.

kevincianfarini: Yeah. The only useful piece of information from a district office is its phone number. If an office doesn't have one, then it's useless to us. 

Therefore we just mark to column not null and delete all of the null rows. 

Thanks for double checking me! :smile:

PR #321, opened 2022-05-04
Add GPL

kevincianfarini:


EvanHahn (approved):
If you're asking whether you successfully included the GPL, I think you did! (Signal, licensed under the AGPL, puts a copyright notice at the top of every file. _Maybe_ necessary here, but probably not...I am not a lawyer.)

If you're asking for feedback on which license to choose, I wonder if a more conservative license (like the [ACSL](https://anticapitalist.software/)) might avoid people using this code for Bad Things. But GPL is fine, too!

In any case, this PR looks good.

kevincianfarini:
@EvanHahn This is mainly done because Gitlab is requiring that open source organizations have a valid open source license in order to get free unlimited CI minutes. I just chose GPL because it's the most copyleft license I know of. I'll have to look into ACSL, but for the most part I don't really have a strong opinion.

PR #324, opened 2022-05-16
Move away from deprecated apis as they are now error

hmit:
These are deprecated with an ERROR level now causing the builds to fail.


kevincianfarini:
@hmit deprecation errors don't start surfacing until you upgrade versions. Have you forgotten to bump a version somewhere in this PR? We're also seeing build failure on CI indicating that OkHttp doesn't know what some of these references are.

hmit:
@kevincianfarini yep, I figured out what's going on. I added slack api as a dep and that bumped okhttp3 underneath. I didn't include that change in this PR thinking it shouldn't cause any problems. If you are okay, I can bump the okhttp version in this PR (or include that in my PR).

kevincianfarini:
That comment has me under the impression that you're making a PR into this repo for the slack bot. Is that correct? 

If so, can we split it into a separate repo? This repo is specifically for the action tool.

kevincianfarini:
We can always bump okhttp versions and migrate accoedingly here, too. ☺️

Issue #295, opened 2022-02-14
`member_of_congress` should have `first_name` and `last_name` 


kevincianfarini:
These names won't really be used besides for the occasional data sort.

PR #329, opened 2022-05-19
Add first name and last name to `member_of_congress`

kevincianfarini:
References #295

### DIFF
--- a/backend/src/main/sqldelight/migrations/V8__add_MoC_first_last_name.sqm
+++ b/backend/src/main/sqldelight/migrations/V8__add_MoC_first_last_name.sqm
@@ -1,0 +1,3 @@
+ALTER TABLE member_of_congress
+ADD COLUMN first_name VARCHAR,
+ADD COLUMN last_name VARCHAR;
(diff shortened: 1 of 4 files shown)


backend/src/main/sqldelight/migrations/V8__add_MoC_first_last_name.sqm
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE member_of_congress
+ADD COLUMN first_name VARCHAR,
+ADD COLUMN last_name VARCHAR;
```
kevincianfarini, line 3: This is backwards compatible. We will make this `NOT NULL` in a future migration.

backend/src/test/kotlin/org/climatechangemakers/act/feature/util/SqlDriverExtensions.kt
```diff
@@ -25,7 +25,22 @@ fun SqlDriver.insertIssue(
 }
 
 fun SqlDriver.insertMemberOfCongress(member: MemberOfCongress) = execute(
-  0, "INSERT INTO member_of_congress VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?)", 9
+  identifier = 0,
```
EvanHahn, line 28: I don't think this query is any different, other than being easier to read. Lemme know if I'm missing something.

EvanHahn (approved):

PR #330, opened 2022-05-20
Use latest member of congress parser

### LINKED ISSUE #295: `member_of_congress` should have `first_name` and `last_name`

kevincianfarini:
Closes #295


congress/legislators-current.json
```diff
@@ -50798,4 +50798,4 @@
       }
     ]
   }
-]
\ No newline at end of file
+]
```
kevincianfarini, line 50801: to trigger the correct CI jobs. Consider coming up with a more robust solution. Manual, maybe?

EvanHahn (approved):

Issue #323, opened 2022-05-16
Consider using directed acyclic graph for CI pipeline


kevincianfarini:
Our pipeline is starting to get pretty complex. We should optimize it for speed.

PR #331, opened 2022-05-21
Use latest parse-congress-info

kevincianfarini:
The prior commit only updated the district office CI jobs by accident.
This commit properly applies the updates parser everywhere.


kevincianfarini:
I accidentally fucked up the CI pipeline in #330. It lends credence to the necessity of #323.

PR #338, opened 2022-05-29
Process historical Members of Congress

kevincianfarini:
The United State Congress Legislators Github project will remove a
legislator from the `legislators-current.json` file if they resign and
move them to the `legislators-historical.json` file.

We need this information if Members of Congress resign outside of the
election cycle, like Rep. Delgado did this week.

PR #339, opened 2022-05-30
Only select currently serving Members of Congress

kevincianfarini:
Now that we're tracking more Members of Congress than are currently
serving, we need to use `member_of_congress.term_end` to determine if a
given member is currently serving.

We should only be selecting MoC who are currently serving, as they're
the people who are meant to receive correspondence.

### DIFF
--- a/backend/build.gradle.kts
+++ b/backend/build.gradle.kts
@@ -54,6 +54,8 @@ dependencies {
 
     implementation(libs.logback.classic)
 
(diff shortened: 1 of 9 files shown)


kevincianfarini:
This unblocks #338

kevincianfarini:
I'm going to eagerly merge. I've got some downtime right now! Will follow up review comments.

EvanHahn:
Code LGTM.

PR #341, opened 2022-06-24
Update Members of Congress

kevincianfarini:


congress/active-offices.json
```diff
@@ -1,166 +1,166 @@
 [
-    {
```
kevincianfarini, line 2: This is the result of `jq` in the update script. There are no changes here.

congress/update.sh
```diff
@@ -0,0 +1,7 @@
+!#/bin/bash
```
EvanHahn, line 1: Do we want those `set -e` flags here and so on?

EvanHahn (approved):

PR #361, opened 2022-10-30
Stub out basic HTTP authentication

kevincianfarini:


EvanHahn (commented):
I admit I don't have a great understanding here. Is this just scaffolding, or should I be seeing some stubbing/mocking?

kevincianfarini:
@EvanHahn this is just scaffolding. 

I might switch from http basic auth to http digest auth so that a password isn't being ferried around between client and server.

EvanHahn (approved):

PR #362, opened 2022-11-02
Verify user credentials for authenticated routes

kevincianfarini:
This commit includes a database migration which creates a new user
table. The credentials held within this table, namely the password, are
hashed with bcrypt.


kevincianfarini:
I should almost certainly add a salt to this. I misunderstood the value-add of salts.

backend/src/main/kotlin/org/climatechangemakers/act/feature/cms/manager/DatabaseUserVerificationManager.kt
```diff
@@ -0,0 +1,28 @@
...
+) : UserVerificationManager {
+
+  private val contentManagementUserQueries = database.contentManagementUserQueries
+
+  override suspend fun verifyLogin(
+    username: String,
+    password: String,
+  ): Boolean = withContext(ioContext) {
+    // Use the username as a salt below by concatenating the two values together.
+    val passwordHash = ByteString.encodeUtf8("$username$password").sha512().hex()
```
kevincianfarini, line 22: @EvanHahn With my understanding of salts, as long as it's globally unique (which username _is_), then it should suffice? Is there any other added benefit to using a random sample of data for the salt that I'm missing?

kevincianfarini: Ugh, if usernames can ever change (which we shouldn't rule out) they can't be used as a salt. Fixing...

kevincianfarini: Working on using Bcrypt for this instead of SHA-512. It's a bit more nuanced.

kevincianfarini:
@EvanHahn This is ready for review!

EvanHahn:
I'll plan to look this weekend.

backend/gradle/libs.versions.toml
```diff
@@ -14,40 +15,31 @@ testcontainers = "1.16.3"
 xmlutil = "0.84.1"
 
 [libraries]
+bcrypt = { module = "at.favre.lib:bcrypt", version.ref = "bcrypt" }
```
EvanHahn, line 18: Trusting that these are just reordered with `bcrypt` added.

EvanHahn (approved):

backend/gradle/libs.versions.toml
kevincianfarini: That's exactly right!

PR #381, opened 2022-12-17
Initial frontend cms

jacknugent:
image

Done:
- Display bills/issues
- Update bills/issues
- Create bills
- Creating issues
- Deleting issues/bills
- Adding talking points to issues

Next up:
- Connect to deployment (initial setup just like CCA frontend)
- Get feedback for table/modal content/order of properties
- Get feedback for layout


frontend_cms/README.md
```diff
@@ -0,0 +1,46 @@
+# Getting Started with Create React App
```
kevincianfarini, line 1: Can we delete this file?

frontend_cms/public/robots.txt
```diff
@@ -0,0 +1,3 @@
+# https://www.robotstxt.org/robotstxt.html
+User-agent: *
+Disallow:
```
kevincianfarini, line 3: I don't think we want to allow scraping here. Can we disallow everything?

frontend_cms/src/Components/CongressionalBills/CongressionalBillsModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+const BILL_TYPES = [
+    "H.R.",
+    "H.Con.Res.",
+    "H.J.Res.",
+    "H.Res.",
+    "S.",
+    "S.Con.Res.",
+    "S.J.Res.",
+    "S.Res.",
+];
```
kevincianfarini, line 17: These should be fetched from the Bill API. If we don't have one yet, I should add it.

frontend_cms/README.md
jacknugent: Hmm we could but I also think it's kind of helpful. Would be better to have CCA specific README but I think this generated one is better than nothing

kevincianfarini: CCA specific would be good!

frontend_cms/src/Components/CongressionalBills/CongressionalBillsModal.tsx
jacknugent: Unless I'm mistaken there isn't one yet

frontend_cms/public/robots.txt
jacknugent: Added `/` to robots.txt which disallows scraping

frontend_cms/README.md
jacknugent: Done

frontend_cms/src/Components/CongressionalBills/CongressionalBillsModal.tsx
kevincianfarini: I will add it.

kevincianfarini: Resolved. #387

kevincianfarini:
This looks like a great first go. I have a few small nitpicks, but they can be addressed a bit later. 

1. The display state of a bill should be the propert bill name/function. Eg. "H.R. 2307 The Energy Innovation and Carbon Dividend Act of 2021". Right now I think it's using the database ID and looks like "1 - Energy Innovation and Carbon Dividend Act of 2021"
2. It would be great if the update modal for both bills and issue would be wider. This is especially important for drafting talking points within issues. 

I think those are my only comments for now!

kevincianfarini (approved):

PR #394, opened 2023-03-08
Null coalesce `relatedBillIds` to empty list

kevincianfarini:


kevincianfarini:
Going to merge so this fix is available for Gabrielle.

